PR #15, opened 2025-06-27
[APG-1041] Remove unused HMPPS domain events topic configuration from Helm values

pfcairns:
* Remove spurious namespace secret that is causing deployment failure


Jack-Dempster:
are we not going to need this topic to subscribe to our events?

Jack-Dempster (approved):

PR #21, opened 2025-07-02
[APG-1012] Add support for referral entity, referral status history, and related persistence logic

pfcairns:
- Created `ReferralEntity` and `ReferralStatusHistoryEntity` with mappings
- Added corresponding repositories and factories for testing
- Extended `MessageHistoryEntity` to include optional referral association
- Updated tests and integration tests for referral handling
- Added database migrations for referral-related tables
- Updated Docker Compose to include PostgreSQL and LocalStack configurations for testing

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
@@ -10,7 +10,7 @@ import jakarta.persistence.JoinColumn
 import jakarta.persistence.JoinTable
 import jakarta.persistence.OneToMany
 import jakarta.persistence.Table
-import java.time.LocalDate
+import org.springframework.data.annotation.CreatedDate
 import java.time.LocalDateTime
 import java.util.UUID
 
@@ -23,18 +23,13 @@ class ReferralEntity(
   var id: UUID? = null,
 
   @Column(name = "person_name")
-  var personName: String? = null,
+  var personName: String,
 
   @Column(name = "crn")
-  var crn: String? = null,
-
-  @Column(name = "cohort")
-  var cohort: String? = null,
-
-  @Column(name = "sentence_end_date")
-  var sentenceEndDate: LocalDate? = null,
+  var crn: String,
 
   @Column(name = "created_at")
+  @CreatedDate
   var createdAt: LocalDateTime = LocalDateTime.now(),
 
   @OneToMany(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
@@ -1,0 +1,51 @@
+package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.entity
+
+import jakarta.persistence.CascadeType
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.FetchType
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.JoinTable
+import jakarta.persistence.OneToMany
+import jakarta.persistence.Table
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.util.UUID
+
+@Entity
+@Table(name = "referral")
+class ReferralEntity(
+  @Id
+  @GeneratedValue
+  @Column(name = "id")
+  var id: UUID? = null,
+
+  @Column(name = "person_name")
+  var personName: String? = null,
+
+  @Column(name = "crn")
+  var crn: String? = null,
+
+  @Column(name = "cohort")
+  var cohort: String? = null,
+
+  @Column(name = "sentence_end_date")
+  var sentenceEndDate: LocalDate? = null,
+
+  @Column(name = "created_at")
+  var createdAt: LocalDateTime = LocalDateTime.now(),
+
+  @OneToMany(
+    fetch = FetchType.LAZY,
+    cascade = [CascadeType.ALL],
+    orphanRemoval = true,
+  )
+  @JoinTable(
+    name = "referral_status_history_mapping",
+    joinColumns = [JoinColumn(name = "referral_id")],
+    inverseJoinColumns = [JoinColumn(name = "referral_status_history_id")],
+  )
+  var statusHistories: MutableList<ReferralStatusHistoryEntity> = mutableListOf(),
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralStatusHistoryEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralStatusHistoryEntity.kt
@@ -1,0 +1,33 @@
+package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.entity
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.time.LocalDateTime
+import java.util.*
+
+@Entity
+@Table(name = "referral_status_history")
+class ReferralStatusHistoryEntity(
+  @Id
+  @GeneratedValue
+  @Column(name = "id")
+  var id: UUID? = null,
+
+  @Column(name = "status")
+  var status: String? = null,
+
+  @Column(name = "created_at")
+  var createdAt: LocalDateTime = LocalDateTime.now(),
+
+  @Column(name = "created_by")
+  var createdBy: String? = null,
+
+  @Column(name = "start_date")
+  var startDate: LocalDateTime? = null,
+
+  @Column(name = "end_date")
+  var endDate: LocalDateTime? = null,
+)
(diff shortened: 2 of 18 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
```diff
@@ -0,0 +1,51 @@
...
+  var id: UUID? = null,
+
+  @Column(name = "person_name")
+  var personName: String? = null,
+
+  @Column(name = "crn")
+  var crn: String? = null,
+
+  @Column(name = "cohort")
+  var cohort: String? = null,
```
thomaswilsonxyz, line 32: can we strip these out for now, we'll add them in as/when we need them

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
```diff
@@ -0,0 +1,51 @@
...
+
+  @Column(name = "sentence_end_date")
+  var sentenceEndDate: LocalDate? = null,
```
thomaswilsonxyz, line 35: same here, can we remove - we'll add them in when we get more info about them

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
```diff
@@ -0,0 +1,51 @@
...
+    fetch = FetchType.LAZY,
+    cascade = [CascadeType.ALL],
+    orphanRemoval = true,
+  )
+  @JoinTable(
+    name = "referral_status_history_mapping",
+    joinColumns = [JoinColumn(name = "referral_id")],
+    inverseJoinColumns = [JoinColumn(name = "referral_status_history_id")],
+  )
+  var statusHistories: MutableList<ReferralStatusHistoryEntity> = mutableListOf(),
```
thomaswilsonxyz, line 50: do we need to do anything here e.g. filter on `deleted_at=null` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralStatusHistoryEntity.kt
```diff
@@ -0,0 +1,33 @@
...
+  var createdAt: LocalDateTime = LocalDateTime.now(),
+
+  @Column(name = "created_by")
+  var createdBy: String? = null,
+
+  @Column(name = "start_date")
+  var startDate: LocalDateTime? = null,
+
+  @Column(name = "end_date")
+  var endDate: LocalDateTime? = null,
```
thomaswilsonxyz, line 32: what do these represent ?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
pfcairns: Sure.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
```diff
@@ -0,0 +1,51 @@
...
+import jakarta.persistence.Table
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.util.UUID
+
+@Entity
+@Table(name = "referral")
+class ReferralEntity(
+  @Id
+  @GeneratedValue
```
Jack-Dempster, line 21: I don't think we should generate an Id here we can use the id provided from `Find and Refer` so they are the same across services potentially?

src/main/resources/db/migration/V2__create_referral_tables.sql
```diff
@@ -0,0 +1,27 @@
...
+    crn TEXT,
+    cohort TEXT,
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE TABLE IF NOT EXISTS referral_status_history (
+    id UUID NOT NULL PRIMARY KEY,
+    status TEXT NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    created_by TEXT,
```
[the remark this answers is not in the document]
pfcairns: It would more than likely be a username or user id so text is more appropriate.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/listener/ReferralCreatedHandler.kt
```diff
@@ -12,7 +12,7 @@ class ReferralCreatedHandler(
 
 ) {
   fun handle(domainEventMessage: DomainEventsMessage) {
-    val sqsMessageHistoryEntity = DomainEventsMessage.toEntity(domainEventMessage, objectMapper)
-    messageHistoryRepository.save(sqsMessageHistoryEntity)
+    val messageHistoryEntity = DomainEventsMessage.toEntity(domainEventMessage, objectMapper)
```
Jack-Dempster, line 15: can we just read the string here rather than passing the objectMapper in?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
pfcairns: To get the current status, we would need to filter on `entity.endDate != null`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralStatusHistoryEntity.kt
pfcairns: Yes @Jack-Dempster, there are `startDate` and and and `endDate` columns for this purpose.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
Jack-Dempster: What do you mean by an already used ID? Events should be idempotent in theory so we should get some sort of exception if we are looking to create a referral with the same id. If we don't store the find and refer Id at all then we have no way of knowing whether we are creating duplicate referrals unless we start looking at the details of that referral.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralStatusHistoryEntity.kt
Jack-Dempster: sorry I worded that terribly. I was wondering what it would look like it we didn't have the endData column at all

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralStatusHistoryEntity.kt
```diff
@@ -0,0 +1,33 @@
...
+@Entity
+@Table(name = "referral_status_history")
+class ReferralStatusHistoryEntity(
+  @Id
+  @GeneratedValue
+  @Column(name = "id")
+  var id: UUID? = null,
+
+  @Column(name = "status")
+  var status: String? = null,
```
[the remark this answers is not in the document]
Jack-Dempster: Not sure I understand this one. In Find we have been able to use non-null fields in the entities https://github.com/ministryofjustice/hmpps-find-and-refer-an-intervention-service/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/findandreferanintervention/jpa/entity/Referral.kt

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/listener/ReferralCreatedHandler.kt
```diff
@@ -12,7 +12,7 @@ class ReferralCreatedHandler(
...
+    messageHistoryRepository.save(messageHistoryEntity)
```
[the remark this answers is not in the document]
Jack-Dempster: I like using extension methods here but it's just personal preference. Since in your code it would just be `domainEventsMessage.toEntity()` and no need to pass anything in

`fun Referral.toDto(): ReferralDetailsDto = ReferralDetailsDto(
  interventionType = this.interventionType,
  interventionName = this.interventionName,
  personReference = this.personReference,
  personReferenceType = this.personReferenceType,
  referralId = this.id,
  setting = this.settingType,
  sourcedFromReference = this.sourcedFromReference,
  sourcedFromReferenceType = this.sourcedFromReferenceType,
)`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/listener/ReferralCreatedHandler.kt
pfcairns: Yeah, I didn't really want to pass it in either tbh. I'll see if i can come up with an alternative.

src/main/resources/db/migration/V2__create_referral_tables.sql
```diff
@@ -0,0 +1,27 @@
...
+    sentence_end_date DATE,
+    start_date TIMESTAMP,
```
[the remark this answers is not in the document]
pfcairns: Given we will want to set this manually when a status changes, probably not.

PR #46, opened 2025-07-17
[APG-1168] Add endpoint to create availability

anandelangovan1:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityController.kt
```diff
@@ -7,18 +7,22 @@ import io.swagger.v3.oas.annotations.media.Schema
...
+import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestMethod
 import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.ErrorResponse
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.model.Availability
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.service.AvailabilityService
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.service.toModel
 import java.util.UUID
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.model.create.AvailabilityCreate as AvailablityCreateModel
```
pfcairns, line 23: This Kotlin feature is horrendous IMHO  as it obfuscates the actual implementation class.   Case in point I had to go and look for the import to see what this Model was.  It also smells of the classic code smell: 

https://medium.com/@codechuckle/code-quality-unleashed-the-principle-of-least-astonishment-ac805cb9bd40

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/AvailabilityEntity.kt
```diff
@@ -35,7 +36,7 @@ class AvailabilityEntity(
   val otherDetails: String? = null,
 
   @Column(name = "last_modified_by", nullable = false)
-  var lastModifiedBy: String,
+  var lastModifiedBy: String = SecurityContextHolder.getContext().authentication?.name!!,
```
pfcairns, line 39: This feels like an odd default value to have in an entity. I'm wondering if there is another way in JPA to automatically set this using an annotation. Maybe not for now.

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/integration/IntegrationTestBase.kt
```diff
@@ -142,4 +142,22 @@ abstract class IntegrationTestBase {
     .expectStatus().isEqualTo(expectedResponseStatus)
     .expectBody(returnType)
     .returnResult().responseBody!!
+
+  fun <T> performRequestAndExpectStatusWithBody(
```
pfcairns, line 146: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/model/create/CreateAvailability.kt
```diff
@@ -0,0 +1,29 @@
...
+import java.util.UUID
+
+data class CreateAvailability(
+
+  @get:JsonProperty("referralId", required = true)
+  @Schema(example = "d3f55f38-7c7b-4b6e-9aa1-e7d7f9e3e785", description = "The ID of the referral", required = true)
+  val referralId: UUID,
+
+  @get:JsonProperty("startDate")
+  @Schema(example = "2025-07-10", description = "Start date of the availability")
```
thomaswilsonxyz, line 16: could we mention if this defaults to `now` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/model/create/CreateAvailability.kt
```diff
@@ -0,0 +1,29 @@
...
+
+  @get:JsonProperty("endDate")
+  @Schema(example = "2025-07-20", description = "End date of the availability", nullable = true)
+  val endDate: LocalDateTime? = null,
+
+  @get:JsonProperty("otherDetails")
+  @Schema(example = "Available for remote sessions", description = "Additional details", nullable = true)
+  val otherDetails: String? = null,
+
+  @JsonProperty("availability")
```
thomaswilsonxyz, line 27: why the mismatch of `availability` to `availabilities` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/AvailabilityEntity.kt
```diff
@@ -35,6 +36,7 @@ class AvailabilityEntity(
   val otherDetails: String? = null,
 
   @Column(name = "last_modified_by", nullable = false)
+  @LastModifiedBy
   var lastModifiedBy: String,
```
thomaswilsonxyz, line 40: can we add a doc string here to just specify that the string `"UNKNOWN"` is a possible value if the system couldn't set it effectively from the barer token

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/AvailabilityTransformer.kt
```diff
@@ -0,0 +1,75 @@
...
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+fun CreateAvailability.toEntity(): AvailabilityEntity {
+  val availabilityEntity = AvailabilityEntity(
+    referralId = this.referralId,
+    startDate = this.startDate?.toLocalDate() ?: LocalDate.now(),
+    endDate = this.endDate?.toLocalDate(),
+    otherDetails = this.otherDetails,
+    lastModifiedBy = SecurityContextHolder.getContext().authentication?.name ?: "UNKNOWN",
```
thomaswilsonxyz, line 21: I think it's probably worth popping this value as an argument to this function, because calling `toEntity` shouldn't have an implicit dependency on the `SecurityContextHolder` in the Spring Book environment

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
```diff
@@ -79,4 +126,40 @@ class AvailabilityControllerIntegrationTest : IntegrationTestBase() {
       )
     }
   }
+
+  fun buildAvailabilityCreateModel(
```
thomaswilsonxyz, line 130: does this factory need to be extracted out to be used in other tests ?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/DefaultAvailabilityConfigService.kt
```diff
@@ -31,3 +31,16 @@ fun DayOfWeek.toPluralLabel(): String = when (this) {
   DayOfWeek.SATURDAY -> "Saturdays"
   DayOfWeek.SUNDAY -> "Sundays"
 }
+
+fun String.toDayOfWeek(): DayOfWeek = when (this) {
```
thomaswilsonxyz, line 35: i don't think this belongs on `String`, which is a pretty foundational/abstract namespace to put this on. 

I think instead we should try putting it on the `DailyAvailabilityModel`, and perhaps name it something like `labelToDayofWeek`.

If we're going to do that, then I reckon we could even add a `.dayOfWeek` property to the `DailyAvailabilityModel` which is set based on the `label` value, which would sort of remove the need for any public method like this, because we would just need to access e.g. `availability.dayOfWeek` 

That would encapsulate this logic where it is relevant, give us the same observed behaviour, and keep our global name spaces free of utilities which are (presently) only used in one place.

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
```diff
@@ -68,6 +71,50 @@ class AvailabilityControllerIntegrationTest : IntegrationTestBase() {
...
+      expectedResponseStatus = HttpStatus.CREATED.value(),
+    )
+
+    assertThat(availability.id.toString()).isNotNull
+    assertThat(availability.referralId.toString()).isEqualTo(referralEntity.id.toString())
+    assertThat(availability.startDate).isNotNull
+    assertThat(availability.endDate).isNotNull
+    assertThat(availability.otherDetails).isEqualTo("Available remotely")
+    assertThat(availability.lastModifiedBy).isEqualTo("AUTH_ADM")
+    assertThat(availability.lastModifiedAt).isNotNull
```
thomaswilsonxyz, line 93: I don't love that all of these test the default behaviour of the `ReferralEntityFactory`, I know it's more verbose, but could we set the values in the POST request explicitly to show that the two are actually connected together.

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
```diff
@@ -68,6 +71,50 @@ class AvailabilityControllerIntegrationTest : IntegrationTestBase() {
...
+
+    assertThat(availability.availabilities).hasSize(7)
+
+    val monday = availability.availabilities.find { it.label == "Mondays" }
```
thomaswilsonxyz, line 97: Yeah, seeing this magic string (`"Mondays"`), even though it's a test, makes me double-down on adding some greater detail to the `AvailabilityEntity`, and not having the `label` be a string with some magic known values.

It would, e.g., be much cleaner to have `it.dayOfWeek.equals(DayOfWeek.Monday)`

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/AvailabilityTransformerTest.kt
```diff
@@ -0,0 +1,79 @@
...
+          label = "Wednesdays",
+          slots = listOf(
+            Slot(label = "daytime", value = false),
+            Slot(label = "evening", value = true),
+          ),
+        ),
+      ),
+    )
+
+    mockkStatic(SecurityContextHolder::class)
```
thomaswilsonxyz, line 49: this would be able to go away if we injected the user id as an argument, which would make this test cleaner, which feels like confirmation that it's a good idea

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/model/create/CreateAvailability.kt
anandelangovan1: I wanted to have the flexibility of mentioning a start date in the future. Do you want me to specify it defaults to now?

anandelangovan1: Done, added to createAvailability model

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
anandelangovan1: For now, this is the only class which uses this. Probably something to consider when this needs to be used elsewhere

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/model/Availability.kt
```diff
@@ -49,8 +50,29 @@ data class Slot(
...
+
+enum class AvailabilityOption(val displayName: String) {
+
+  MONDAY("Mondays"),
+  TUESDAY("Tuesdays"),
+  WEDNESDAY("Wednesdays"),
+  THURSDAY("Thursdays"),
+  FRIDAY("Fridays"),
+  SATURDAY("Saturday"),
+  SUNDAY("Sunday"),
```
Jack-Dempster, line 67: should this be `Saturdays` and `Sundays`

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
anandelangovan1: Addressed based on previous conversation. 
Switched to using enums

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/model/Availability.kt
anandelangovan1: Good spot, addressed and pushed

Jack-Dempster (dismissed):
Approved from my side of things

PR #62, opened 2025-07-25
[APG-1062] Update availability to take a date instead of timestamp

anandelangovan1:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/AvailabilityService.kt
```diff
@@ -38,10 +38,16 @@ class AvailabilityService(
...
+    if (existingAvailability != null) {
+      log.info("Availability already exists for referralId ${createAvailability.referralId}")
+      return Pair(existingAvailability.toModel(), true)
+    }
+
+    val availabilityEntity = createAvailability.toEntity(getAuthenticatedReferrerUser())
     val savedAvailabilityEntity = availabilityRepository.save(availabilityEntity)
-    return savedAvailabilityEntity.toModel()
+    return Pair(savedAvailabilityEntity.toModel(), false)
   }
```
thomaswilsonxyz, line 51: the return type `Pair<Availability, Boolean>` doesn't feel especially clear or readable, specifically what that `Boolean` represents.  In general we should aim for method names and / or documentation that makes it clear what's being returned without requiring someone knows the implementation.

Could we either: 

- Add a human-friendly doc string on the method to indicate what `true`/`false` represents
- Update the function name to make it clear what the `true`/`false` represents`
- (possible but less favourable) Add a named field in the return type, e.g. ` isNew: Boolean`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/AvailabilityTransformer.kt
```diff
@@ -16,8 +16,8 @@ import java.time.LocalDateTime
 fun CreateAvailability.toEntity(lastModifiedBy: String = SecurityContextHolder.getContext().authentication?.name ?: "UNKNOWN"): AvailabilityEntity {
   val availabilityEntity = AvailabilityEntity(
     referralId = this.referralId,
-    startDate = this.startDate?.toLocalDate() ?: LocalDate.now(),
-    endDate = this.endDate?.toLocalDate(),
+    startDate = this.startDate ?: LocalDate.now(),
```
thomaswilsonxyz, line 19: I don't think we should be defaulting the `startDate` field here in the `toEntity` context

To my mind, this bit of code is about transforming us from one entity to another(a `CreateAvailability`) into another (`AvailabilityEntity`).  It doesn't feel right to _also_ mutate state as we make that transition.

A more accurate name would be something like `:toEntityWithDefaultValues(lastModifiedBy...)` which reads funny.

src/main/resources/db/migration/V9__update_availability_table.sql
```diff
@@ -0,0 +1,25 @@
...
+    start_date        DATE NOT NULL,
+    end_date          DATE,
+    other_details     TEXT,
+    last_modified_by  VARCHAR NOT NULL,
+    last_modified_at  TIMESTAMP NOT NULL,
+    CONSTRAINT fk_referral FOREIGN KEY (referral_id) REFERENCES referral (id)
+);
+
+
+CREATE TABLE slot
```
thomaswilsonxyz, line 18: oooh, could we possibly call this table something like `availability_slot` - to help provide a bit more context to the humans who work around the database.  The word `slot` could also come to mean something for another user of our system (e.g. group setup for Training Managers) so by keeping our database terms a bit more semantically precise, we can avoid confusion at the database level, without having to go into the Kotlin code to double-check was entity `slot` is referring to

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
```diff
@@ -164,17 +164,36 @@ class AvailabilityControllerIntegrationTest : IntegrationTestBase() {
...
+    val referralEntity = ReferralEntityFactory().produce()
+    testDataGenerator.createReferral(referralEntity)
+
+    val createdAvailability = createAvailability(referralEntity, startDate, endDate, otherDetails)
+
+    // assert that its a conflict
+    val duplicateAvailability =
+      createAvailability(referralEntity, startDate, endDate, otherDetails, HttpStatus.CONFLICT.value())
+
+    assertThat(createdAvailability.id).isEqualTo(duplicateAvailability.id)
```
thomaswilsonxyz, line 183: assuming you decide to keep the `Pair` architecture, could you add an assertion here on that (e.g. here I _think_ it should be `false`?)

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
```diff
@@ -164,17 +164,36 @@ class AvailabilityControllerIntegrationTest : IntegrationTestBase() {
...
-    val startDate: LocalDateTime = LocalDateTime.now()
-    val endDate: LocalDateTime = LocalDateTime.now().plusDays(10)
+    val startDate: LocalDate = LocalDate.now()
+    val endDate: LocalDate = LocalDate.now().plusDays(10)
 
     val referralEntity = ReferralEntityFactory().produce()
     testDataGenerator.createReferral(referralEntity)
 
-    val createAvailability = createAndGetAvailability(referralEntity, startDate, endDate, otherDetails)
+    val createAvailability = createAvailability(referralEntity, startDate, endDate, otherDetails)
```
thomaswilsonxyz, line 196: similarly, here, can you assert on the `Boolean` value coming back in this Pair ?

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
```diff
@@ -264,11 +283,12 @@ class AvailabilityControllerIntegrationTest : IntegrationTestBase() {
   )
 }
 
-private fun AvailabilityControllerIntegrationTest.createAndGetAvailability(
+private fun AvailabilityControllerIntegrationTest.createAvailability(
   referralEntity: ReferralEntity,
-  startDate: LocalDateTime,
-  endDate: LocalDateTime,
+  startDate: LocalDate,
+  endDate: LocalDate,
```
thomaswilsonxyz, line 289: This is nullable, no ?

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
```diff
@@ -240,8 +259,8 @@ class AvailabilityControllerIntegrationTest : IntegrationTestBase() {
   fun buildUpdateAvailabilityModel(
     availabilityId: UUID,
     referralId: UUID,
-    startDate: LocalDateTime,
-    endDate: LocalDateTime,
+    startDate: LocalDate,
+    endDate: LocalDate,
```
thomaswilsonxyz, line 263: nullable ?

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AvailabilityControllerIntegrationTest.kt
```diff
@@ -283,13 +303,13 @@ private fun AvailabilityControllerIntegrationTest.createAndGetAvailability(
...
-  expectedResponseStatus = HttpStatus.CREATED.value(),
+  expectedResponseStatus = responseCode,
 )
 
 fun buildCreateAvailabilityModel(
   referralId: UUID = UUID.randomUUID(),
-  startDate: LocalDateTime? = LocalDateTime.now().minusDays(1),
-  endDate: LocalDateTime? = LocalDateTime.now().plusDays(10),
+  startDate: LocalDate? = LocalDate.now().minusDays(1),
+  endDate: LocalDate? = LocalDate.now().plusDays(10),
```
thomaswilsonxyz, line 312: does this mean it's not possible to create a `AvailabilityModel` in a test that has `endDate = null` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/AvailabilityService.kt
anandelangovan1: Went with the doc approach.

rgforsyth (approved):

PR #78, opened 2025-07-31
Apg 1156/risk info

ian-moj:


Jack-Dempster (commented):
Looks good, would we be able to add an entry into the wiremock_mappings so we get a response when calling locally.

Jack-Dempster (changes requested):
can we add a wiremock mapping so we can call use this endpoint locally

ian-moj:
Closing as this integration is not needed

PR #83, opened 2025-08-04
Update find and refer details model to accept sourcedFromReference

Jack-Dempster:
This PR updates the Find and Refer referral details model to accept the `sourcedFromReference` field we need to store in Manage and Deliver.

It also enables connection to the dev instances of Find and Refer and NDelius for local development so we do not have to mock these services any more.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
@@ -13,9 +13,9 @@ import jakarta.persistence.JoinTable
 import jakarta.persistence.OneToMany
 import jakarta.persistence.Table
 import jakarta.validation.constraints.NotNull
-import org.hibernate.annotations.JdbcType
-import org.hibernate.dialect.PostgreSQLEnumJdbcType
 import org.springframework.data.annotation.CreatedDate
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.entity.type.InterventionType
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.entity.type.SettingType
 import java.time.LocalDateTime
 import java.util.UUID
 
@@ -37,7 +37,6 @@ class ReferralEntity(
   @NotNull
   @Column(name = "intervention_type")
   @Enumerated(EnumType.STRING)
-  @JdbcType(PostgreSQLEnumJdbcType::class)
   var interventionType: InterventionType,
 
   @Column(name = "intervention_name")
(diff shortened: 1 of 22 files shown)


src/main/resources/db/local/V99_1__add_local_referral_data.sql
```diff
@@ -1,16 +1,16 @@
-insert into referral (id, person_name, crn, created_at, intervention_type, intervention_name, setting)
+insert into referral (id, person_name, crn, created_at, intervention_type, intervention_name, setting, event_number)
 values ('39fde7e8-d2e3-472b-8364-5848bf673aa6', 'Edgar Schiller', 'X718250', '2025-07-30 14:51:22.086567', 'ACP',
-        'Building Choices', 'COMMUNITY'),
+        'Building Choices', 'COMMUNITY', '2500828798'),
```
thomaswilsonxyz, line 3: it's okay that these are all from the same `event_number` ?

src/main/resources/db/migration/V10__Add_event_number_column_to_referral_table.sql
```diff
@@ -0,0 +1,7 @@
+DELETE
```
thomaswilsonxyz, line 1: we sure we want to delete every row in the `referral` table ?

thomaswilsonxyz (approved):

pfcairns (commented):
Looks good. Just a few minor comments.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
```diff
@@ -54,4 +64,32 @@ class ReferralEntity(
     inverseJoinColumns = [JoinColumn(name = "referral_status_history_id")],
   )
   var statusHistories: MutableList<ReferralStatusHistoryEntity> = mutableListOf(),
+
+  @NotNull
+  @Column("eventNumber")
+  val eventNumber: String,
 )
+
+enum class PersonReferenceType {
```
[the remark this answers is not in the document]
Jack-Dempster: yeah good shout I'll move these out into their own package

src/main/resources/db/migration/V10__Add_event_number_column_to_referral_table.sql
pfcairns: Agree. We definitely don't want to do this.

src/main/resources/db/local/V99_1__add_local_referral_data.sql
Jack-Dempster: it won't cause any problems but this is just laziness from me when I was testing so will update them now

src/main/resources/db/migration/V10__Add_event_number_column_to_referral_table.sql
Jack-Dempster: After having thought of this overnight, I've removed the delete and set the default. Although I am expecting them to never be used as we currently do not have any data in our environments.

PR #91, opened 2025-08-08
[APG-1122] Return cohort as part of case list view and add filtering …

anandelangovan1:
…based on cohort


src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/CaseListControllerIntegrationTest.kt
```diff
@@ -26,6 +27,9 @@ class CaseListControllerIntegrationTest : IntegrationTestBase() {
   @Autowired
   private lateinit var resourceLoader: ResourceLoader
 
+  @Autowired
+  private lateinit var testDataGenerator: TestDataGenerator
+
```
Jack-Dempster, line 32: is this used?

Jack-Dempster (approved):
Looks good!

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/CaseListControllerIntegrationTest.kt
Jack-Dempster: Yeah Me and Phil tried to sort this out but for whatever reason the view is never populated with data when using EntityManager. We thought it's probably something to do with the transaction. It's really frustrating but we couldn't figure it out either

PR #97, opened 2025-08-11
Update cohort response to string instead of enum

anandelangovan1:


pfcairns (commented):
Just one question : why? This feels like a backwards step to me.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/CaseListController.kt
```diff
@@ -52,18 +52,22 @@ class CaseListController(private val referralCaseListItemService: ReferralCaseLi
     @Parameter(description = "Filter by the cohort of the referral Eg: SEXUAL_OFFENCE or GENERAL_OFFENCE") @RequestParam(
       value = "cohort",
       required = false,
-    ) cohort: OffenceCohort?,
+    ) cohort: String?,
```
thomaswilsonxyz, line 55: I think you can use the enum here, not a string, as you've got the display values for the enum wired in

thomaswilsonxyz (dismissed):

PR #100, opened 2025-08-12
Handle null event number when trying to retrieve sentence information

Jack-Dempster:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReferralController.kt
```diff
@@ -194,6 +198,10 @@ class ReferralController(
     @PathVariable("id") id: UUID,
   ): ResponseEntity<SentenceInformation> {
     val referral = referralService.getReferralById(id) ?: throw NotFoundException("Referral with id $id not found")
+    if (referral.eventNumber == null) {
+      log.info("Referral with id $id has null eventNumber")
```
thomaswilsonxyz, line 202: we should probably error/warn here, if we're about to `throw` below

`.info` feels like it's "good to know"

thomaswilsonxyz (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReferralController.kt
Jack-Dempster: yeah good spot

thomaswilsonxyz (approved):

PR #119, opened 2025-08-20
Apg 1229/restrict lao

ian-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralCaseListItemService.kt
```diff
@@ -19,6 +25,21 @@ class ReferralCaseListItemService(private val referralCaseListItemRepository: Re
...
+    val filtered = allItems
+      .filter { it.crn in allowedCrns }
+      .map { it.toApi() }
+
+    // apply paging AFTER filtering
+    val start = pageable.offset.toInt()
+    val end = (start + pageable.pageSize).coerceAtMost(filtered.size)
+    val pageContent = if (start <= end) filtered.subList(start, end) else emptyList()
+
+    return PageImpl(pageContent, pageable, filtered.size.toLong())
```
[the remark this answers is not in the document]
thomaswilsonxyz: this actually appears to be a good suggestion from the LLM, and not one that we thought of 

by fetching _just_ the CRNs of all the Referrals that match the search criteria, then checking those, and then using the CRNs to exclude results later on - we are going to the database twice, but we're not over-fetching data.

I'd imagine we'd want to make sure we have an index on the CRN for `referrals`, but my SQL knowledge on indexes isn't amazing

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralCaseListItemService.kt
```diff
@@ -19,6 +26,24 @@ class ReferralCaseListItemService(private val referralCaseListItemRepository: Re
     status: String?,
   ): Page<ReferralCaseListItem> {
     val specification = getReferralCaseListItemSpecification(openOrClosed, crnOrPersonName, cohort, status)
-    return referralCaseListItemRepository.findAll(specification, pageable).map { it.toApi() }
+    val allItems = referralCaseListItemRepository.findAll(specification) // List<Entity>
```
thomaswilsonxyz, line 29: Is is possible to have a `findAllCrns` method here, and _just_ return the CRNs, as opposed to the whole object ?  we can probably increase performance by just working on one column

pfcairns (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralCaseListItemService.kt
ian-moj: Actioned

PR #134, opened 2025-08-28
APG-1019 Add SchemaSpy

KirilKolevMoJ:


Jack-Dempster:
HTML image of artifact being produced:
image

Full generated report available below:
[Database Schema Report.zip](https://github.com/user-attachments/files/22116678/Database.Schema.Report.zip)

PR #139, opened 2025-09-01
APG-1170: Add the sourced_from field to the Referral

thomaswilsonxyz:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReferralEntity.kt
```diff
@@ -70,12 +70,23 @@ class ReferralEntity(
...
   var statusHistories: MutableList<ReferralStatusHistoryEntity> = mutableListOf(),
 
+  @Nullable
+  @Column("sourced_from")
+  var sourcedFrom: ReferralEntitySourcedFrom? = null,
+
+  //  This is an alias to the sourced_from_id, i.e. the Requirement or Licence ID
   @Nullable
   @Column("event_id")
+  @Enumerated(EnumType.STRING)
```
pfcairns, line 80: This annotation needs to be on the sourcedFrom column field instead

PR #148, opened 2025-09-09
Update the DeliveryLocationPreferences data model

thomaswilsonxyz:
## WHAT ?

This PR introduces a set of changes related to the DeliveryLocationPreferences.

Those changes are particularly well documented inside the following doc files: 

- [2025-09-09-removing-region-pdu-office-reference-data.md](docs/schema-docs/2025-09-09-removing-region-pdu-office-reference-data.md)
- [2025-09-09-preferred-delivery-locations.md](docs/schema-docs/2025-09-09-preferred-delivery-locations.md)

This work started as a mobbing session between TJWC, RF, and JD - all three have context on the decisions made

## WHY

Our understanding of how, where, and when Delivery Locations can be sourced has changed in the last week, and we are now retrieving them from nDelius.  Our system therefore needs to be able to persist some idea of "PreferredDeliveryLocations", i.e. geographic places where a Person on Probation can attend the sessions for an Accredited Programme.

The previous iteration (using reference data stored in our own database) has proven non-viable for a handful of reasons, and therefore these changes reflect our current best understanding: that the Manager of the Licence Condition or Requirement which caused the Referral to an Accredited Programme is actually the best source of a list of Offices.  And that list of Offices is associated with a subset of Probation Delivery Units in the real world, which may be (for various pragmatic reasons) duplicates of PDUs already in the data set.  For example, there may be a "North East PDU" and a "North East PDU (Accredited Programmes)", so that offices which cannot/do not deliver Accredited Programmes are not given as possible locations to attend

## Commit Log

- **migration: Remove the office, pdu, region tables**
- **migrations: Create the new database structure for the delivery location preferences**
- **fix: remove the DELETE_ME.md file, used for pairing**


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/PreferredDeliveryLocation.kt
```diff
@@ -0,0 +1,48 @@
...
+  @Column(name = "id")
+  val id: UUID,
+
+  @Column(name = "delius_code")
+  val deliusCode: String,
+
+  @Column(name = "delius_description")
+  val deliusDescription: String,
+
+  @Nonnull
```
Jack-Dempster, line 37:
```suggestion
  @NotNull
```
to match the normal notNull annotation

PR #151, opened 2025-09-10
APG-1112: Create the DeliveryLocationPreference Form BFF endpoint

thomaswilsonxyz:
## What

This PR introduces a new API endpoint at `GET /bff/referral-delivery-location-preferences-form/{referral_id}`, which returns all of the data necessary to display the multi-page form to submit (or update) DeliveryLocationPreferences against a Referral.

It adds OpenAPI / Swagger docs detailing this endpoint.

## Why

We are using the BFF pattern because we need to fetch data from a variety of places (multiple nDelius endpoints, and our own database), and this moves the complexity of the multi-step fetch into the API allowing the UI to remain simple, 

## How

The PR creates a new Service and Controller for the DeliveryLocationPreferences, using standard SpringBoot patterns.  There is nothing interesting or controversial here.

Using the refactored ReferralService (see below) to fetch both the manager and possible delivery locations is the main source of complexity.

Tests with various stubs have also been implemented, with a preference for not stubbing out data.

## Refactoring

This PR also contains some moderate strength refactoring to the ReferralService, which checks to see if a Referral has a relevant Requirement or Licence Condition (and also updates the entity in the database).

This needed extracting because the same flow is used for both the `.manager` and the `.probationDeliveryUnits` at the root of that `/case/{crn}/requirement|licence-condition/{event_id}`

The PR introduces a handful of tests to make sure this flows.

## Commits

- **refactor(nDelius Stubs) Add stubPersonalDetailsResponseForCrn helper method**
- **refactor(ReferralService): Extract various checker/helper methods to check the presence or absence of Requirements or Licence Conditions**
- **refactor(DLP Repository tests): Rename a variable probationDeliveryUnit -> PDU**
- **feature(DeliveryLocationPreferences) Create the GET /bff/referral-delivery-location-preferences-form/{referral_id} endpoint**

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/DeliveryLocationPreferencesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/DeliveryLocationPreferencesController.kt
@@ -1,0 +1,69 @@
+package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.controller
+
+import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.Parameter
+import io.swagger.v3.oas.annotations.media.Content
+import io.swagger.v3.oas.annotations.media.Schema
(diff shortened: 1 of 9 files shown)


Jack-Dempster (approved):
👍

PR #164, opened 2025-09-16
APG-1179: Add LDC history db tables and entities

Jack-Dempster:
This pr adds the `referral_ldc_history` table and related entities.

This enables us to retrieve the ldc status history for a referral.

image


src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/integration/wiremock/HmppsAuthMockServer.kt
```diff
@@ -38,7 +38,7 @@ class HmppsAuthApiExtension :
 
 class HmppsAuthMockServer : WireMockServer(WIREMOCK_PORT) {
   companion object {
-    private const val WIREMOCK_PORT = 8090
+    private const val WIREMOCK_PORT = 9999
```
Jack-Dempster, line 41: this is to avoid conflicts when running tests while the docker containers are running in the background during local dev

src/main/resources/db/migration/V26__create_referral_ldc_history_table.sql
```diff
@@ -0,0 +1,18 @@
...
+
+COMMENT ON TABLE referral_ldc_history IS 'Contains a history of the changes to the LDC (Learning Difficulties and Challenges) for a referral.';
+COMMENT ON COLUMN referral_ldc_history.id IS 'The unique identifier for a referral_ldc_history entry.';
+COMMENT ON COLUMN referral_ldc_history.referral_id IS 'Foreign key relationship to the referral table.';
+COMMENT ON COLUMN referral_ldc_history.has_ldc IS 'If a referral has Learning Difficulties or Challenges.';
+COMMENT ON COLUMN referral_ldc_history.created_by IS 'Username of the person/system that changed the LDC status of the referral';
+COMMENT ON COLUMN referral_ldc_history.created_at IS 'Timestamp of when the LDC status was changed for a referral.';
+
+CREATE INDEX idx_referral_ldc_history_referral_id ON referral_ldc_history (referral_id);
+CREATE INDEX idx_referral_ldc_history_created_by ON referral_ldc_history (created_by);
```
pfcairns, line 18: Do we really need an index on the the `created_by` column?

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/integration/wiremock/HmppsAuthMockServer.kt
pfcairns: you may additionaly need to update the values in the `application-test.yml`

pfcairns (commented):
Should there be en `endDate` field and `end_date` column for the history table? And perhaps a start date too?

Alternatively and as much as I understand why we don't want the `referral` table going down a similar route to its equivalent in  custody, and turning into a god object/table, this could simply be represented as just a boolean value on a referral.

Jack-Dempster:
> Should there be en `endDate` field and `end_date` column for the history table? And perhaps a start date too?
> 
> Alternatively and as much as I understand why we don't want the `referral` table going down a similar route to its equivalent in custody, and turning into a god object/table, this could simply be represented as just a boolean value on a referral.

@pfcairns  we want to capture the history and who changed it for audit reasons but also what if in the future we want to add a description as to why they've changed ldc. Would that then have to be stored on the referral too?

pfcairns (approved):

thomaswilsonxyz:
just my 2c on the `end_date`

I'm wary of adding another way for things to go wrong.  e.g. what happens if the `end_date` of one record doesn't line up with the `start_date` of the next ? should our system note this, or respond ?

even if it's a 0.01% chance of happening, it will happen eventually

in general I'd lean towards having the smallest surface possible to achieve what we need - so if having both `start_date` and  `end_date` can do something that another implementation cannot - then we should absolutely consider it, but I've not encountered anything in this conversation that's made me think that.

extremely happy to be disagreed with and proven wrong or not careful enough on this one

pfcairns (approved):

PR #174, opened 2025-09-19
APG-1182: Add POST endpoint to update LDC status for a referral

Jack-Dempster:
## WHAT

* Adds a POST endpoint for Updating LDC status for a referral

## HOW

* When a request to update ldc status comes through and `hasLdc` property in the body is not null then add entry into LDC history table and with LDC status True or False.


pfcairns (approved):
All good.

PR #185, opened 2025-09-24
[APG-1343] Store referral with default values when PNI doens't exist …

anandelangovan1:
…for a crn


Jack-Dempster:
Looks good but can you also add a test into the `DomainEventsListener` test so we can be sure the event handling works as expected for our referral creation

PR #191, opened 2025-09-24
fix: Set the default createdBy field for a ReferralStatusHistory to SYSTEM, not test_data

thomaswilsonxyz:


Jack-Dempster (approved):
Good spot 👍

PR #195, opened 2025-09-26
[APG-1174] Adding appinsights metrics when an event is received

anandelangovan1:


KirilKolevMoJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/listener/ReferralCreatedHandler.kt
```diff
@@ -26,9 +29,22 @@ class ReferralCreatedHandler(
...
+    }
     val referralId = extractReferralId(domainEventMessage.detailUrl)
     log.info("Received referral created event for referral id: $referralId")
+
+    telemetryClient.logToAppInsights(
+      "Probation.case-requirement.created event received",
+      mapOf(
+        "eventType" to domainEventMessage.eventType,
+        "referralId" to referralId.toString(),
+        "crn" to domainEventMessage.personReference.findCrn()!!,
```
Jack-Dempster, line 44: maybe add event number(?) in here too? I can't remember what the actual field is called

Jack-Dempster: I've just looked and what we send from Find and Refer and we don't actually send the event number across it's fetched at a later date. Sorry, resolving now

PR #202, opened 2025-10-01
APG-1378: Adds the count of the opposite referrals tab to the bff filters

Jack-Dempster:
In order for the ui to display the number of referrals in the other tab than it is currently in we need to provide a count for the `otherReferrals`. This means if the UI sends a request where it is in the `Open` tab we should count the `Closed` referrals and send this number back.

```json
{
    "statusFilters": {
        "open": [
            "Awaiting allocation",
            "Awaiting assessment",
            "Breach (non-attendance)",
            "Deferred",
            "Deprioritised",
            "On programme",
            "Recall",
            "Return to court",
            "Scheduled",
            "Suitable but not ready"
        ],
        "closed": [
            "Programme complete",
            "Withdrawn"
        ]
    },
    "otherReferralsCount": 6
}
```


niaevans9 (approved):
lgtm

PR #203, opened 2025-10-02
APG-1381: Add referral_reporting_location table and entities

Jack-Dempster:


src/main/resources/db/migration/V36__create_referral_reporting_location.sql
```diff
@@ -0,0 +1,61 @@
...
+CREATE INDEX idx_referral_reporting_location_reporting_team ON referral_reporting_location (reporting_team);
+
+
+-- Recreate view and add PDU and Reporting team
+DROP MATERIALIZED VIEW referral_caselist_item_view;
+
+CREATE MATERIALIZED VIEW referral_caselist_item_view AS
+WITH latest_status AS (SELECT referral_id,
+                              referral_status_description_id,
+                              ROW_NUMBER() OVER (PARTITION BY referral_id ORDER BY created_at DESC) as rn
```
Jack-Dempster, line 24: this has been changed from `start_date` -> `created_at` as this was causing an error when returning the latest status. Do we even need this `start_date` column?

src/main/resources/db/migration/V36__create_referral_reporting_location.sql
```diff
@@ -0,0 +1,61 @@
...
+SELECT r.id,
+       r.crn,
+       r.person_name,
+       r.cohort,
+       rsd.description_text                                   as status,
+       -- Default to false if there are no entries in the referral_ldc_history_table
+       COALESCE(lds.has_ldc, false)                           as has_ldc,
+       -- Default values if there are no entries in the referral_reporting_location table for this referral yet
+       COALESCE(rrl.pdu_name, 'UNKNOWN_PDU_NAME')             as pdu_name,
+       COALESCE(rrl.reporting_team, 'UNKNOWN_REPORTING_TEAM') as reporting_team
```
Jack-Dempster, line 40: This is because we may not have entries in the `referral_reporting_location` table and if that's so we still want to display them in the view. This can be changed if we would rather force them to have a value through a migration instead?

PR #523, opened 2026-03-17
APG-1609 : Refactor attendance and attendee relationships so flow is …

Jack-Dempster:
…session -> attendee -> attendance"


Jack-Dempster:
Closing for now, realistically this will probably never be implemented now but would definitely be a nice to have

PR #558, opened 2026-04-09
Apg 2040/edit group days times endpoint

niaevans9:
New endpoint to return Days and Times data for a group so that it can be edited on the UI.
Also return the Days and Times day in day order, Monday first through to Sunday


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
```diff
@@ -304,14 +305,16 @@ class ProgrammeGroupService(
...
-    val daysAndTimes: List<String> = programmeGroup.programmeGroupSessionSlots.map {
-      "${it.dayOfWeek.toAvailabilityOptions()}, ${
-        formatTimeOfSession(
-          it.startTime,
-          it.startTime.plusMinutes(150),
-        )
-      }"
-    }
+    val daysAndTimes: List<String> = programmeGroup.programmeGroupSessionSlots
+      .sortedBy { it.dayOfWeek.value } // Sort by day of week (Monday=1, Sunday=7)
```
raby, line 309: (plus line 415)

Please add a secondary sort by start time; sorting only by day is non-deterministic for multiple slots on the same day since the source is a set.

      .sortedWith(compareBy({ it.dayOfWeek.value }, { it.startTime })) // Sort by day, then time

AI helping explain: 
"The current code sorts session slots only by day of week. Because the underlying collection is a set, entries on the same day do not have a stable order, so two Tuesday slots can appear in different sequences across runs. That can cause inconsistent API responses, flaky tests, and UI jitter. Sorting by day and then start time makes the output deterministic and predictable for consumers."

niaevans9: You can only have a single slot per day

raby: Thats fine then I was overthinking possibilities thank you

rgforsyth (approved):

raby (commented):
Looks great

PR #568, opened 2026-04-14
APG-1968 Add `ldcScore` to LearningNeeds model and update related logic

pfcairns:
### Changes
- Added `ldcScore` field to the LearningNeeds model.
- Updated service logic to handle the new `ldcScore` attribute.
- Extended integration tests to include scenarios for `ldcScore`.


raby (approved):
LGTM

PR #571, opened 2026-04-15
[APG-2110] add updatedAt attribute to attendee

miroslav25a:


src/main/resources/db/migration/V99__add_updated_at_to_attendee.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE attendee
+    ADD COLUMN updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
```
raby: Although self explanatory in this case,  other alter table sqls also add a comment describing the column. Good for consistency?

miroslav25a: comment addressed.

raby (approved):
Looks good

PR #577, opened 2026-04-21
Drop data import table that was created in public schema as it is cau…

Jack-Dempster:
…sing duplicated references in data importer service


raby (approved):
LGTM

PR #580, opened 2026-04-22
Dont include catch up sessions in the list on the schedule page

niaevans9:


src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupServiceIntegrationTest.kt
```diff
@@ -675,6 +675,48 @@ class ProgrammeGroupServiceIntegrationTest : IntegrationTestBase() {
...
+      // When
+      val schedule = service.getScheduleOverviewForGroup(programmeGroup.id!!)
+
+      // Then - only the non-catch-up session should be returned
+      assertThat(schedule.sessions).hasSize(1)
+      assertThat(schedule.sessions.first().date).isEqualTo(LocalDate.of(2026, 7, 20))
+    }
+
     @Test
     fun `should throw NotFoundException if group does not exist`() {
```
raby, line 721: Test Coverage: Catch-up-only Sessions

Just a suggestion for testing all scenarios - even if not strictly possible with how we intend to use the system/data, only having catch-ups I think.

We could add a test for the edge case where a group has sessions, but **all are catch-ups**. After the filter, no sessions remain, so it should throw `NotFoundException`. 

It ensures the error is caught at the right place (empty filtered list) rather than silently returning an incomplete schedule. This follows the same pattern as should throw NotFoundException when group has no sessions, but ensures the error is triggered by filtering rather than a genuinely empty group.

### Example test (following existing patterns):

```kotlin
@Test
fun `should throw NotFoundException when all sessions are catch-ups`() {
  // Given - a group with only catch-up sessions
  testDataGenerator.createSession(
    SessionFactory()
      .withProgrammeGroup(programmeGroup)
      .withModuleSessionTemplate(regularModuleSessions.first { it.sessionType == SessionType.GROUP })
      .withStartsAt(LocalDateTime.of(2026, 7, 20, 15, 30))
      .withEndsAt(LocalDateTime.of(2026, 7, 20, 17, 30))
      .withIsPlaceholder(false)
      .withIsCatchup(true)
      .produce(),
  )
  testDataGenerator.createSession(
    SessionFactory()
      .withProgrammeGroup(programmeGroup)
      .withModuleSessionTemplate(regularModuleSessions.first { it.sessionType == SessionType.ONE_TO_ONE })
      .withStartsAt(LocalDateTime.of(2026, 7, 27, 15, 30))
      .withEndsAt(LocalDateTime.of(2026, 7, 27, 17, 30))
      .withIsPlaceholder(false)
      .withIsCatchup(true)
      .produce(),
  )

  // When/Then - all sessions are filtered out, so NotFoundException is thrown
  val exception = assertThrows<NotFoundException> {
    service.getScheduleOverviewForGroup(programmeGroup.id!!)
  }

  assertThat(exception.message).isEqualTo("No sessions found for group ${programmeGroup.id}")
}

Comment (niaevans9) [reply src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupServiceIntegrationTest.kt:721]: The tests are slow enough to run as is I dont think we should be clogging them up with scenarios that will never occur

File: src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt [hunk 2]
```diff
@@ -589,6 +589,7 @@ class ProgrammeGroupService(
 
     val groupSessions =
       sessions.filter { it.sessionType == SessionType.GROUP || (it.sessionType == SessionType.ONE_TO_ONE && it.isPlaceholder) }
```
Jack-Dempster: just stick the filter in here too so we aren't iterating over the list twice
```suggestion
      sessions.filter { (it.sessionType == SessionType.GROUP && !it.isCatchup) || (it.sessionType == SessionType.ONE_TO_ONE && it.isPlaceholder && !it.isCatchup)  }
```

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
```diff
@@ -588,7 +588,8 @@ class ProgrammeGroupService(
     }
 
     val groupSessions =
-      sessions.filter { it.sessionType == SessionType.GROUP || (it.sessionType == SessionType.ONE_TO_ONE && it.isPlaceholder) }
+      sessions.filter { (it.sessionType == SessionType.GROUP && !it.isCatchup) || (it.sessionType == SessionType.ONE_TO_ONE && it.isPlaceholder && !it.isCatchup) }
+        .filter { !it.isCatchup }
```
Jack-Dempster:
```suggestion

```

Jack-Dempster (approved):

PR #585, opened 2026-04-23
Apg 1993/dont reschedule pre group one to one

niaevans9:
Change to the logic around editing a group start date
If you edit the group start date
Even if the group start date does not match the cadence of the group (e.g slots are on Mondays at 1pm and you change the start date to be a Friday)
Regardless of if the user chooses to automatically reschedule sessions
The pre-group one-to-one session placeholder start date should be set to the start date selected when editing a group


Jack-Dempster (approved):
👍

PR #589, opened 2026-04-24
Return the whole facilitator object instead of just the name

niaevans9:
UI requires the whole facilitator object instead of just the name, also formatted so its in the correct form for the UI to use easily


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/model/programmeGroup/GroupDetailsResponse.kt
```diff
@@ -104,25 +104,25 @@ data class GroupDetailsResponse(
   val currentlyAllocatedNumber: Int,
 
   @Schema(
-    example = "Chloe Pascal",
+    example = "{\"personCode\": \"CP001\", \"personName\": \"Chloe Pascal\", \"teamName\": \"Management Team\", \"teamCode\": \"TEAM001\"}",
```
Jack-Dempster, line 107: just for future reference if you set examples on the actual `UserTeamMember` class on the individual properties the you don't need to set this crazy json string in this response class

Jack-Dempster (approved):
added a comment but this doesn't stop merging, just a future improvement 👍

raby (approved):

Jack-Dempster (approved):

raby (approved):

PR #590, opened 2026-04-24
APG-2177 Add error handling for updating appointments in delius

Jack-Dempster:
When updating appointments in delius we were missing any exception handling being returned from the client


niaevans9 (approved):
lgtm

raby (commented):
LGTM

PR #600, opened 2026-04-29
only call ndelius for future apopintments when allocating to a group created in the past.

rgforsyth:
When creating a group in the past, so only some sessions are in the future, only call nDelius to create the sessions in the future.


Jack-Dempster (commented):
👍

Jack-Dempster (approved):

raby (commented):
LGTM

PR #601, opened 2026-04-29
Update the remove from group success message

niaevans9:
Update the success message when removing someone from a group to be in the format:
'[Alex River] was removed from this group. Their referral status is now [status, eg Return to court]'.


pfcairns (approved):
LGTM.

PR #604, opened 2026-04-29
Return sessions on the schedule page in date/time order

niaevans9:
Return sessions on the schedule page in date/start time order
If it is a placeholder session and has times of various set the start time to be midnight so it appears first for that day

If a user did then move a session out of the usual order (eg moving Getting started 2 earlier than Getting started 1), that would be reflected in the schedule overview order, which would help the user to see that they’d done something odd.


DanEgan1 (approved):

raby (commented):
Looks good!

PR #611, opened 2026-05-05
APG-2136 Update session rescheduling logic 

pfcairns:
- Session rescheduling now handles. and additionally reschedules placeholder one-to-one and catch-up sessions distinctly
- Updated related tests, and fixed inconsistent naming in tests and test data generation


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
```diff
@@ -124,12 +124,12 @@ class SessionService(
...
     val isGroupSession = session.sessionType == SessionType.GROUP
-    val shouldShiftOtherGroupSessions = request.rescheduleOtherSessions && isGroupSession
 
-    if (shouldShiftOtherGroupSessions) {
+    // Reschedule group sessions, place-holder one-to-ones, but not catch-up sessions
+    if (request.rescheduleOtherSessions) {
       val subsequentGroupSessions = session.programmeGroup.sessions
         .asSequence()
-        .filter { it.sessionType == SessionType.GROUP }
+        .filter { !it.isCatchup }
```
raby, line 137: Great progress catching the catch-up case here! One more exclusion needed though — the ticket specifies that scheduled one-to-ones (non-placeholder actual sessions) should also stay fixed. The current filter !it.isCatchup will pick those up and move them unintentionally? Or am I missing something :) 

`ProgrammeGroupService.getScheduleOverviewForGroup` uses `.filter { (it.sessionType == SessionType.GROUP || it.isPlaceholder) && !it.isCatchup }` could that help?

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/SessionControllerIntegrationTest.kt
```diff
@@ -469,17 +508,22 @@ class SessionControllerIntegrationTest : IntegrationTestBase() {
...
-    // Session 3 should NOT be moved as it is INDIVIDUAL
-    val updatedSession3 = sessionRepository.findById(session3.id!!).get()
-    assertThat(updatedSession3.startsAt).isEqualTo(LocalDateTime.of(2026, 4, 25, 10, 0))
-    assertThat(updatedSession3.endsAt).isEqualTo(LocalDateTime.of(2026, 4, 25, 11, 0))
+    // Session 3 should be rescheduled it is ONE_TO_ONE place holder session
+    val updatedPlaceholderOneToOneSession = sessionRepository.findById(placeholderOneToOneSession.id!!).get()
+    assertThat(updatedPlaceholderOneToOneSession.startsAt).isEqualTo(LocalDateTime.of(2026, 4, 25, 11, 0))
+    assertThat(updatedPlaceholderOneToOneSession.endsAt).isEqualTo(LocalDateTime.of(2026, 4, 25, 12, 0))
+
+    // Catch up sessions shouldn't be rescheduled
```
raby: Good shape on the new test scenarios — the placeholder move and catch-up freeze are both well covered, nice one.

Would it be worth adding one more session here: a non-placeholder one-to-one after the edited session, with an assertion that it does not move? That would give protection for all three rules from the ticket.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
raby: Line 128 comment

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
```diff
@@ -124,12 +124,12 @@ class SessionService(
...
         .filter { it.startsAt.isAfter(session.startsAt) }
```
raby: Nit: session.startsAt has already been mutated at line 123 before this filter runs, so the comparison is against the new start time rather than the original. 

In normal usage (shifting within the same day) this is fine, but if a session were ever moved forward by several days it could silently skip sessions that fall between the old and new datetime. 

A quick fix would be to capture `val originalStartsAt = session.startsAt` before line 123 and filter against that instead possibly?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
pfcairns: Good spot Raby. Will address.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
pfcairns: Yep - I've shifted the date that we filter on to be the incoming non-mutated `session.startsAt.`

Jack-Dempster (approved):
👍 Looks good! Test setup looks painful!

raby (approved):
Looks good

PR #614, opened 2026-05-06
Apg 2210/same date for 2 sessions bug

miroslav25a:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
```diff
@@ -161,36 +161,8 @@ class ProgrammeGroupService(
...
-          )
-
-          sessionService.rescheduleSessions(
-            preGroupOneToOnePlaceholderSession.id!!,
-            rescheduleRequest,
-          )
-
-          log.info("Successfully rescheduled pre-group one-to-one placeholder session ${preGroupOneToOnePlaceholderSession.id} for group $groupId")
-        }
+      if (updateGroupRequest.automaticallyRescheduleOtherSessions == false) {
```
raby: Nice clean extraction of reschedulePreGroupOneToOneSession 👍

One thing I was thinking about: the old code always rescheduled the placeholder whenever earliestStartDate changed. With this change, if automaticallyRescheduleOtherSessions is omitted (i.e. null in the api contract rather then front-end use), neither branch it enters — so the DB may have an updated earliestPossibleStartDate but the placeholder session would still be on the old date?

Since the current frontend always sends an explicit true or false, this probably isn't a practical concern right now. But it might be worth either:

doing `!= true` instead of `== false `so null falls into the same branch as false?

Or documenting the null-means-no-operation behaviour with a test

Happy to merge either way — just a minor api contract possibly

pfcairns (dismissed):
LGTM

raby (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
miroslav25a: Good point, will address the comment.

miroslav25a: comment addressed.

pfcairns (approved):

Jack-Dempster (approved):
👍

raby (approved):
Great

PR #616, opened 2026-05-06
update description

Alison-Weatherston:
Returning the names of the participants rather than 'All' for more than one.

Updated example text

 example = "[\"John Doe\", \"Jane Smith\"]")

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
@@ -439,7 +439,7 @@ class ProgrammeGroupService(
               scheduledSession.startsAt.toLocalTime(),
               scheduledSession.endsAt.toLocalTime(),
             ),
-            participants = if (sessionTemplate.sessionType == SessionType.GROUP) listOf("All") else scheduledSession.attendees.map { it.personName },
+            participants = scheduledSession.attendees.map { it.personName },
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
```diff
@@ -452,7 +452,7 @@ class ProgrammeGroupService(
               scheduledSession.startsAt.toLocalTime(),
               scheduledSession.endsAt.toLocalTime(),
             ),
-            participants = if (sessionTemplate.sessionType == SessionType.GROUP) listOf("All") else scheduledSession.attendees.map { it.personName },
+            participants = scheduledSession.attendees.map { it.personName },
```
Jack-Dempster: Are we not still supposed to return `All` for group sessions? It's just all catchup and individual sessions should give the names of the people?

Jack-Dempster (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
Alison-Weatherston: Thanks @Jack-Dempster - it was showing 'All' for group catchups too, which it is not supposed to. I was initially going to differentiate it in the UI, but I have done it in the API after your comment.

Jack-Dempster: great thanks for sorting, generally easier to do these kind of things in the api anyway

Jack-Dempster (approved):
👍

PR #624, opened 2026-05-08
APG-2238 : Ensure LAO check on caselist is limited to 500 crns max

Jack-Dempster:
## WHAT

On the caselist we make a call to check if the logged in user is permitted to access the crns that are being displayed. The current implementation looks at ALL the crns in the db which in higher environments is 1000's. The endpoint which accepts a list of CRNs to check only accepts max 500 crns.

This PR limits the amount of crns sent to the access check endpoint to be 500.


miroslav25a (approved):

raby (commented):
Looks good

PR #628, opened 2026-05-11
feat: Automatically increment eventNumber from 0..20 for imported Referrals

thomaswilsonxyz:
**WHAT**

Created the `ReferralEventNumberResolverService`, and connected it to the `/referral-details/{id}/personal-details` route.

The `resolveIfEventNumberIsZero` method attempts to resolve the `eventNumber` all the way to `20` if it starts at 0 (i.e. if it was imported from the Data Importer Service)

Added custom event logging to AppInsights (as well as regular .info and .warn logging)

**WHY**

When data is imported from the Interventions Manager (IM) backup, we set the eventNumber to 0, because (so far) it appears that the IM data does not contain the eventNumber.

However, when we reach out to get any information about an offence, we require both the licence/requirement ID (which we *do* have) and the eventNumber.  Without it, nDelius/OASys return 4xx errors.

As a naieve solution to this problem, we can attempt to simply iterate up, one at a time, with Referrals.  We have been doing this manually as part of the data import checking process, and it appears to work (though is time consuming)

**HOW**

Introduced a new, single-purpose, service (`ReferralEventNumberResolutionService`) and plugged this into a single endpoint.

This implementation method allows for relatively easy extraction/removal, should a more sophisticated solution become available (e.g. an import from nDelius), and also allows configuration on a per-endpoint basis.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
@@ -1,0 +1,114 @@
+package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.service
+
+import com.microsoft.applicationinsights.TelemetryClient
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.ClientResult
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.nDeliusIntegrationApi.NDeliusIntegrationApiClient
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.config.logToAppInsights
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.entity.ReferralEntity
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.repository.ReferralRepository
+
+private const val FIRST_VALID_EVENT_NUMBER = 1
+private const val LAST_VALID_EVENT_NUMBER = 20
+
+@Service
+class ReferralEventNumberResolverService(
+  private val nDeliusIntegrationApiClient: NDeliusIntegrationApiClient,
+  private val referralRepository: ReferralRepository,
+  private val telemetryClient: TelemetryClient,
+) {
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  /**
+   * When Referrals are populated by the data-importer-service, they have the eventNumber of 0, we know that this
+   * is an invalid number, and so when we attempt to fetch offence details associated with it, we will always get
+   * a 4xx error.  It does not appear that the eventNumber is stored in Interventions Manager data, and therefore
+   * we need a naieve solution which simply:
+   * 1. Recognises that an event number is 0
+   * 2. Iterates up, one at a time (until 20), to see if we can get a 2xx response from nDelius
+   * 3. If no success is found, we leave the event number as 0
+   */
+  @Transactional
+  fun resolveIfEventNumberIsZero(referral: ReferralEntity): Int? {
+    if (referral.eventNumber != 0) {
+      return referral.eventNumber
+    }
+
+    log.info(
+      "Referral {} has event number 0. Attempting to resolve by checking nDelius sentence endpoint for CRN {} ({}..{}).",
+      referral.id,
+      referral.crn,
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverServiceTest.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    assertThat(result).isEqualTo(0)
+    assertThat(referral.eventNumber).isEqualTo(0)
+    verify(referralRepository, never()).save(referral)
+    (1..20).forEach { candidate ->
+      verify(nDeliusIntegrationApiClient).getSentenceInformation(referral.crn, candidate)
+    }
+  }
+
+  private fun mockSentenceResponse() = uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.factory.NDeliusSentenceResponseFactory().produce()
```
Jack-Dempster: remove full qualifier here

Jack-Dempster (dismissed):
Looks great! only 1 tiny  comment

Jack-Dempster (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    telemetryClient.logToAppInsights(
+      "Referral.event-number-resolution.failure",
+      mapOf(
+        "referralId" to referral.id.toString(),
+      ),
+    )
+  }
+
+  private fun logSuccess(referral: ReferralEntity, newEventNumber: Int) {
```
raby, line 98: Small nit: Currently the first valid event number is 1 so they're the same number as newEventNumber, but if the first valid event number changes to, say, 5, and the first match is event number 5, the log would say "after 5 attempt(s)" when it was actually 1 attempt.

```
private fun logSuccess(referral: ReferralEntity, newEventNumber: Int, attemptCount: Int) {
  log.info(
    "Resolved event number for referral {}. Using {} after {} attempt(s).",
    referral.id,
    newEventNumber,
    attemptCount,
  )
  // ...existing telemetry code...
}
```

Line 54 where this is called:

`logSuccess(referral, candidateEventNumber, candidateEventNumber - FIRST_VALID_EVENT_NUMBER + 1)`

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverServiceTest.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.http.HttpMethod
+import org.springframework.http.HttpStatusCode
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.ClientResult
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.nDeliusIntegrationApi.NDeliusIntegrationApiClient
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.factory.NDeliusSentenceResponseFactory
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.factory.ReferralEntityFactory
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.repository.ReferralRepository
+
+@ExtendWith(MockitoExtension::class)
+class ReferralEventNumberResolverServiceTest {
```
raby, line 22: We could add a test testing the failure.other path, I have a working/tested test below helped by an LLM:

```
@Test
fun `continues to next candidate when an unexpected error occurs`() {
  val referral = ReferralEntityFactory().withEventNumber(0).produce()
  val service = ReferralEventNumberResolverService(nDeliusIntegrationApiClient, referralRepository, telemetryClient)

  `when`(nDeliusIntegrationApiClient.getSentenceInformation(referral.crn, 1)).thenReturn(
    ClientResult.Failure.Other(
      HttpMethod.GET,
      "/case/${referral.crn}/sentence/1",
      RuntimeException("Connection refused"),
      "nDelius",
    ),
  )
  `when`(nDeliusIntegrationApiClient.getSentenceInformation(referral.crn, 2)).thenReturn(
    ClientResult.Success(HttpStatusCode.valueOf(200), mockSentenceResponse()),
  )

  val result = service.resolveIfEventNumberIsZero(referral)

  assertThat(result).isEqualTo(2)
  assertThat(referral.eventNumber).isEqualTo(2)
  verify(referralRepository).save(referral)
  verify(nDeliusIntegrationApiClient).getSentenceInformation(referral.crn, 1)
  verify(nDeliusIntegrationApiClient).getSentenceInformation(referral.crn, 2)
}
```

PR #634, opened 2026-05-12
APG-2169 Refactor `ProgrammeGroupService` and query filters for clarity

pfcairns:
This pull request includes the following changes:
- Simplified the logic within `ProgrammeGroupService` by criteria based code to Specification specific classes
- Added criteria filtering for the COMPLETE group tab to include referrals that have a status of Programme Complete, and also ensuring that each referral has an attendance of the final Post programme review session
- Modularized the query filters for region tabs, making them more maintainable and reusable.
- Increased test coverage to verify the refactored components.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/specification/GetProgrammeGroupsSpecification.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/specification/GetProgrammeGroupsSpecification.kt
@@ -136,12 +136,12 @@ fun incompleteMembershipCountSubquery(
   val isMembershipComplete = cb.and(
     isLatestReferralStatusProgrammeComplete(query, cb, referralJoin),
     hasAttendedPostProgrammeReview(incompleteMembershipCountSubquery, cb, groupMembershipRoot),
+    cb.isNotNull(groupMembershipRoot.get<LocalDateTime>("deletedAt")),
   )
 
   incompleteMembershipCountSubquery.select(cb.count(groupMembershipRoot))
   incompleteMembershipCountSubquery.where(
     cb.equal(groupMembershipRoot.get<ProgrammeGroupEntity>("programmeGroup"), root),
-    cb.isNull(groupMembershipRoot.get<LocalDateTime>("deletedAt")),
     cb.not(isMembershipComplete),
   )
   return incompleteMembershipCountSubquery
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/specification/GetProgrammeGroupsSpecification.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/specification/GetProgrammeGroupsSpecification.kt
@@ -1,12 +1,29 @@
 package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.repository.specification
 
+import jakarta.persistence.criteria.CriteriaBuilder
+import jakarta.persistence.criteria.CriteriaQuery
+import jakarta.persistence.criteria.Join
 import jakarta.persistence.criteria.Predicate
+import jakarta.persistence.criteria.Root
+import jakarta.persistence.criteria.Subquery
 import org.springframework.data.jpa.domain.Specification
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.OffenceCohort
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.programmeGroup.ProgrammeGroupCohort
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.type.GroupPageByRegionTab
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.type.ProgrammeGroupSexEnum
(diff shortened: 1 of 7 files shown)


miroslav25a (dismissed):

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ProgrammeGroupControllerIntegrationTest.kt
```diff
@@ -526,6 +542,50 @@ class ProgrammeGroupControllerIntegrationTest : IntegrationTestBase() {
...
+      )
+      unacceptableAbsenceOutcome = sessionAttendanceOutcomeTypeRepository.save(
+        SessionAttendanceNDeliusOutcomeEntity(
+          code = UAAB,
+          description = "Unacceptable Absence",
+          attendance = false,
+          compliant = false,
+        ),
+      )
+    }
```
Jack-Dempster, line 552: are these not already in the db as reference data?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/specification/GetProgrammeGroupsSpecification.kt
```diff
@@ -54,3 +71,106 @@ fun getProgrammeGroupsSpecification(
...
+
+  val isMembershipComplete = cb.and(
+    isLatestReferralStatusProgrammeComplete(query, cb, referralJoin),
+    hasAttendedPostProgrammeReview(incompleteMembershipCountSubquery, cb, groupMembershipRoot),
+  )
+
+  incompleteMembershipCountSubquery.select(cb.count(groupMembershipRoot))
+  incompleteMembershipCountSubquery.where(
+    cb.equal(groupMembershipRoot.get<ProgrammeGroupEntity>("programmeGroup"), root),
+    cb.isNull(groupMembershipRoot.get<LocalDateTime>("deletedAt")),
```
Jack-Dempster: when a referral is moved to `Programme complete` it's removed from a group and `deletedAt` will not be null. Is this going to cause a problem here?

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ProgrammeGroupControllerIntegrationTest.kt
```diff
@@ -565,40 +625,45 @@ class ProgrammeGroupControllerIntegrationTest : IntegrationTestBase() {
...
           referral1,
-          referralStatusDescriptionOnProgramme,
+          onProgrammeReferralStatus,
         ),
       )
       testDataGenerator.creatReferralStatusHistory(
         ReferralStatusHistoryEntityFactory().produce(
           referral1,
-          referralStatusDescriptionCompleted,
+          programmeCompleteReferralStatus,
```
Jack-Dempster, line 634: When this is done through the ui this referral will be removed from the group when they are moved to programmeComplete

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ProgrammeGroupControllerIntegrationTest.kt
```diff
@@ -648,40 +711,44 @@ class ProgrammeGroupControllerIntegrationTest : IntegrationTestBase() {
...
```
Jack-Dempster, line 720: same here the referral will be removed from the group which I don't know if it will cause issues with the counts/tab content

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ProgrammeGroupControllerIntegrationTest.kt
pfcairns: Yes, good call. There is no need to do this, so will pull from the DB instead.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/specification/GetProgrammeGroupsSpecification.kt
pfcairns: I'm not sure the check for the null `deletedAt` value is relevant in this scenario, so will remove it.

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ProgrammeGroupControllerIntegrationTest.kt
pfcairns: I've had a good look at the code, specifically in the `updateStatus` method of the `ReferralService` which then calls `programmeGroupMembershipService.deleteGroupMembershipForReferralAndGroup()`, and it doesn't look like the referral itself is removed from the group. In this scenario which is when the final Post programme review has been attended, there should in theory be no further future sessions for an individual, so the `deleteGroupMembershipForReferralAndGroup` function won't to much bar "soft deleting" the group membership.

Jack-Dempster (approved):
Thanks for making those changes and adding that test 👍

PR #641, opened 2026-05-14
Apg 2271/Fix NoSuchElementException when there are no attendees for a ONE_TO_ONE session

miroslav25a:
Fix NoSuchElementException when there are no attendees for a ONE_TO_ONE session. Currently when a referral is removed from a group all the future one-to-one sessions exist with no attendees


pfcairns (approved):
👍

raby (approved):
LGTM

PR #642, opened 2026-05-14
feat: Create the ReportingGroupSize table view

thomaswilsonxyz:
**WHAT**

This commit introduces a postgresql Materialized View, which brings on
data related to the group size reporting format.

**HOW**

Using a Materialized View allows us to pull through data stored in other
tables, taking a best point-in-time export of data.  It does so in a way
that keeps our data up to date, and is performant (because we are
working at the database level, not the entity level, in-memory with JPA).

The commit introduces a new JPA entity (`ReportingGroupSizeEntitry`) to
model the data in this view, and creates the related Repository that
allows us to fetch this data on two date-based fields.

The commit also introduces some TestDataCleaners to work on the
Materialized View, because they need to be `REFRESH`'d to bring new data
in.

**NEXT STEPS**

Next steps for this work will be creating an endpoint that exposes this
data, and then after that, creating some (hidden) routes on the UI that
can be used to retrieve the data and give it to users.


src/main/resources/db/migration/R__create_reporting_group_size_view.sql
Jack-Dempster, line 1: not entirely necessary but I've found having these views as R__ migrations easier to change and also review when changes are made. It's fine to recreate the view on every deployment I think

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReportingGroupSizeEntity.kt
```diff
@@ -0,0 +1,74 @@
...
+  @Column(name = "created_at")
+  var createdAt: LocalDateTime,
+
+  @NotNull
+  @Column(name = "sex")
+  var sex: String,
+
+  @NotNull
+  @Column(name = "cohort")
+  var cohort: String,
```
Jack-Dempster: can we have this as our cohort enum? Not sure if its `OffenceCohort` or `ProgrammeGroupCohort`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/ReportingGroupSizeEntity.kt
```diff
@@ -0,0 +1,74 @@
...
+  @Column(name = "location_name")
+  var locationName: String,
+
+  @NotNull
+  @Column(name = "group_size")
+  var groupSize: Int,
+
+  @Nullable
+  @Column(name = "facilitator_staff_code")
+  var facilitatorStaffCode: String? = null,
```
Jack-Dempster, line 79: is this nullable?

thomaswilsonxyz: Yeah, from the `ProgrammeGroupEntity`: 

```kt
  @ManyToOne(fetch = FetchType.LAZY)
  @JoinColumn(name = "treatment_manager_id", referencedColumnName = "id")
  var treatmentManager: FacilitatorEntity? = null,
  ```

src/main/resources/db/migration/R__create_reporting_group_size_view.sql
thomaswilsonxyz: oooh, yeah, good shout - til about the Repeatable Migration pattern, thanks !

Jack-Dempster, line 1: also you will need to add triggers to refresh the view. look at group_wait_list_view for examples

rgforsyth (approved):

PR #644, opened 2026-05-14
Adds Sar template and tests.

rgforsyth:
Adds sar template.
Adds hmpps sar testing library
Overwrites tests as appropriate.

Adds custom shell script to regenerate snapshot files, copy and rename them, and them store in the correct folder for the tests to access.

Added readme doc on updating sar tests.


docs/how-to/update-sar-tests.md
```diff
@@ -0,0 +1,171 @@
+# How to Update SAR Tests
```
rgforsyth, line 1: llm generated

scripts/local-scripts/regenerate-sar-snapshots.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
rgforsyth, line 1: Not the default way of updating snapshots but it makes it a lot simpler for devs to update the tests in the future. It runs through all the steps that would need to be manually done otherwise

This script runs the tests with the env var 'SAR_GENERATE_ACTUAL' set to true. This causes snapshots to be created.

It then takes those files, renames and copies them into the correct folder ready for the tests to use them.

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/sar/SarContractIntegrationTest.kt
```diff
@@ -0,0 +1,263 @@
...
+      sarIntegrationTestHelper.saveContentToFile(normalizedActual, "sar-generated-report.html.log")
+    } else {
+      sarIntegrationTestHelper.assertHtmlEquals(
+        normalizedActual,
+        normalizeDynamicValues(sarIntegrationTestHelper.getExpectedRenderResult()),
+      )
+    }
+  }
+
+  private fun normalizeDynamicValues(value: String): String = value
```
rgforsyth, line 249: Doing this when creating the snapshot and upon running the test allows us to not run into timestamp issues/differences between the snapshot and future runs.

raby (approved):
Really great work - had a good deep dive 

Spotted this while reviewing — not introduced here, but worth raising as close-by in the code-base (I may have missed it on a previous review myself!):

src/test/kotlin/.../service/SubjectAccessRequestServiceTest.kt lines 274–281

The referralCaseListItemViews assertions are checking fields from the wrong entity (groupWaitlistItemViewEntity instead of referralCaseListItemViewEntity)? I think? The tests still pass because both factories happen to use identical default values (crn = "X123456", personName = "John Smith" etc.), but this means the referralCaseListItemViews mapping could be completely broken and these assertions would never catch it.

`assertThat(resultContent.referralCaseListItemViews[0].crn).isEqualTo(referralCaseListItemViewEntity.crn)

assertThat(resultContent.referralCaseListItemViews[0].hasLdc).isEqualTo(referralCaseListItemViewEntity.hasLdc)
 assertThat(resultContent.referralCaseListItemViews[0].personName).isEqualTo(referralCaseListItemViewEntity.personName)

assertThat(resultContent.referralCaseListItemViews[0].sentenceEndDate).isEqualTo(referralCaseListItemViewEntity.sentenceEndDate)
 assertThat(resultContent.referralCaseListItemViews[0].status).isEqualTo(referralCaseListItemViewEntity.status)`

Happy to raise a separate ticket if instead or dismiss if I am imagining a bug due to lack of understanding which is possible :-)

PR #646, opened 2026-05-18
APG-2158 Update referral cohort to a history table

Jack-Dempster:
This PR is a pre-requisite to be able to refresh the cohort on the referral details page load. In order to be able to capture when a user has overriden the cohort that is being returned from Oasys we need to add a history table, similar to how the LDC status is tracked, so we can add entries and check if the latest entry is system added or user added.

This has had the side effect of having to change how we retrieve cohort for a referral across the service.

This has been tested on the UI by:
* Updating cohort on referral details page
* Viewing cohort on caselist
* Viewing cohort on group wait list page


miroslav25a (dismissed):

.snyk
```diff
@@ -10,10 +10,5 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
 version: v1.25.1
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
-ignore:
-  SNYK-JAVA-ORGAPACHETOMCATEMBED-16643259:
-    - '*':
-        reason: Only relevant for WebDAV commands
-        expires: 2026-07-01T00:00:00.000Z
-        created: 2026-05-13T09:38:00.250Z
+ignore: {}
```
pfcairns: Should this file be checked into the repo?

pfcairns (dismissed):
Good work. Just the one question, approving in any case.

raby (dismissed):
LGTM

.snyk
Jack-Dempster: I've removed it and added to .gitignore now

pfcairns (approved):

PR #647, opened 2026-05-18
feat: Create reporting/group-size endpoint

thomaswilsonxyz:
**WHAT**

This commit introduces the `GET /reporting/group-size.csv` endpoint.

This returns information about groups (found by their
`earliestPossibleStartDate`) and their size (i.e. current membership).

This is a point-in-time recording, and it is not possible to replay
events, or revert to a certain point in time after the data has been
exported.

The endpoint returns data as a .csv file, serialisation for which is
done with bespoke code.

**WHY**

The design of the current system prioritises speed of delivery of
features.  This is to minimise the time to learning, and maximise the
time to further iterate or develop on the data that the users want.

Once of the four metrics (alongside rate, dosage, and facilitator
continuity) that Performance and Quality (P&Q) teams care about is Group
Size, i.e. how many people are in the Groups.

This endpoint will be called directly by users who want to retrieve the
data.  The current state of reporting is that users are exporting data
from IM, and double-checking it, and therefore we expect there to be a
period of time where we learn about how exactly data is used by the P&Q
teams, and what format (if any) we can present it in which will allow
them to get their jobs done quicker.

**HOW**

This commit introduces a new Controller (`ReportingController`) and
Service (`ReportingService`) which link together with the prexisting
 ReportingGroupSizeRepository to retrieve Group Size data.

The data is serialised into .csv format using in-house code, because I
was not able to find evidence of the kotlinx-serialization-csv library
that the official Kotlin docs link to.  This is unlike the use of
*Jackson* to format data into .json format.

**TESTING**

Integration tests for the Service and the Endpoint.

A (skipped) test that asserts on the user role, we can't implement this
until the role we want to use / create has been instantiated and
assigned to a user.  Otherwise we're going to have an endpoint that
nobody can access


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReportingService.kt
```diff
@@ -0,0 +1,89 @@
...
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.repository.ReportingGroupSizeRepository
+import java.time.ZoneId
+import java.util.Date
+
+@Service
+class ReportingService(
+  private val reportingGroupSizeRepository: ReportingGroupSizeRepository,
+) {
+  companion object {
+    private const val GROUP_SIZE_CSV_HEADER =
```
raby, line 16: Suggestion (non-blocking)

The CSV header is defined as a manual string constant could drift.

Jackson's CsvMapper can generate the header row automatically if you set setUseHeader(true) on the schema, removing the constant and the manual header-prepending logic:

LLM:


`> // Before — two separate sources of truth
> private const val GROUP_SIZE_CSV_HEADER = "id,code,createdAt,..."
>
> private val groupSizeCsvSchema: CsvSchema = CsvSchema.builder()
>   .addColumn("id")
>   .addColumn("code")
>   // ...
>   .build()
>
> // In getGroupSizeReportCsv:
> return if (csvBody.isBlank()) GROUP_SIZE_CSV_HEADER else "$GROUP_SIZE_CSV_HEADER\n$csvBody"`




`> // After — single source of truth
> private val groupSizeCsvSchema: CsvSchema = CsvSchema.builder()
>   .addColumn("id")
>   .addColumn("code")
>   // ...
>   .setUseHeader(true)  // ← Jackson writes the header row automatically
>   .build()
>
> // In getGroupSizeReportCsv — no manual header needed:
> return csvMapper.writer(groupSizeCsvSchema).writeValueAsString(csvRows).trimEnd('\n', '\r')`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReportingController.kt
```diff
@@ -0,0 +1,100 @@
...
+  )
+  @GetMapping("/reporting/group-size.csv", produces = [CSV_MEDIA_TYPE])
+  fun getGroupSizeCsv(
+    @Parameter(
+      description = "Only include groups that started after this date-time (must be in the past).",
+      required = true,
+      example = "2026-05-18T13:30:00",
+    )
+    @RequestParam(name = "groupStartedSince", required = true)
+    @DateTimeFormat(iso = DateTimeFormat.ISO.DATE_TIME)
```
raby, line 83: The controller converts LocalDateTime → Date using ZoneId.systemDefault(), then the service immediately converts Date → LocalDate using ZoneId.systemDefault() again. Two timezone conversions for a value that starts and ends as a local date  and both depend on the JVM's system timezone at runtime, which can differ between environments.

Since the repository just needs a LocalDate, the cleanest fix is to pass LocalDateTime directly through to the service and skip the Date round-trip entirely:

LLM: 

`> // ReportingController.kt — before
> val csv = reportingService.getGroupSizeReportCsv(
>   Date.from(groupStartedSince.atZone(ZoneId.systemDefault()).toInstant()),
> )
>
> // ReportingController.kt — after
> val csv = reportingService.getGroupSizeReportCsv(groupStartedSince)`



`> // ReportingService.kt — before
> fun getGroupSizeReportCsv(firstSessionAfter: Date): String { ... }
>
> private fun getGroupSizeReportRows(firstSessionAfter: Date): List<ReportingGroupSizeEntity> {
>   val firstSessionDate = firstSessionAfter.toInstant().atZone(ZoneId.systemDefault()).toLocalDate()
>   return reportingGroupSizeRepository.getAllGroupsWithEarliestStartDateAfter(firstSessionDate)
> }
>
> // ReportingService.kt — after
> fun getGroupSizeReportCsv(firstSessionAfter: LocalDateTime): String { ... }
>
> private fun getGroupSizeReportRows(firstSessionAfter: LocalDateTime): List<ReportingGroupSizeEntity> {
>   return reportingGroupSizeRepository.getAllGroupsWithEarliestStartDateAfter(firstSessionAfter.toLocalDate())
> }`

rgforsyth (approved):

PR #650, opened 2026-05-18
APG-2229 Set eventNumber for referrals with eventNumber 0 via delius

Jack-Dempster:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
```diff
@@ -39,39 +37,57 @@ class ReferralEventNumberResolverService(
...
-          referralRepository.save(referral)
-          logSuccess(referral, candidateEventNumber)
-          return candidateEventNumber
-        }
+    val eventNumberResponse = when (referral.sourcedFrom) {
+      ReferralEntitySourcedFrom.REQUIREMENT -> {
+        log.info("...attempting to retrieve a Requirement for Referral with ID: ${referral.id}")
+        when (
+          val response =
+            nDeliusIntegrationApiClient.getRequirementManagerDetails(referral.crn, referral.eventId!!)
```
raby: Great improvement using sourcedFrom to target the right endpoint directly — much cleaner than the iterate approach. One small concern here and on line 63...referral.eventId!! will throw a NullPointerException if a referral somehow arrives with sourcedFrom = REQUIREMENT or LICENCE_CONDITION but eventId = null (which isn't impossible for malformed imported data)

Could add a null check before the when block to fail gracefully?

LLM:

`if (referral.eventId == null) {
>   log.warn("Referral with ID '${referral.id}' has no eventId — cannot resolve event number.")
>   logFailureEvent(referral)
>   return referral.eventNumber
> }
> // then use referral.eventId (non-null smart cast) in both branches below`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
```diff
@@ -39,39 +37,57 @@ class ReferralEventNumberResolverService(
...
+          is ClientResult.Success -> response.body
+
+          else -> {
+            log.error("Could not fetch a Licence condition with ID ${referral.eventId}, for Referral with ID: ${referral.id}")
+            null
+          }
         }
       }
+
+      else -> null
```
raby: Non blocking small thing, the else branch on line 74 silently returns null with no log message if sourcedFrom is null or an unexpected value. warning log only says "could not resolve" — which doesn't make it obvious that the issue was an unhandled sourcedFrom rather than an API error. 


`> // Line 74 — current
> else -> null
>
> // Suggested
> else -> {
>   log.warn(
>     "Referral with ID '${referral.id}' has unhandled sourcedFrom value '${referral.sourcedFrom}' — skipping event number resolution.",
>   )
>   null
> }`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
Jack-Dempster: have put it a null check now

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
Jack-Dempster: have added some logging for invalid sourcedFrom now

raby (approved):
Looks great

PR #654, opened 2026-05-19
APG-2158 Automatically refresh cohort on referral details page

Jack-Dempster:
## WHAT

This PR adds the functionality to update the cohort when the `referral-details` endpoint is called. The value will be overwritten if;
* It is different that the current value
* AND it has not been previously set by the user

## NOTES

As part of the implementation of this feature I realised the LDC value was never being updated due to the user always being set as the logged in user rather than our `SYSTEM` value. This has been fixed and also applied to the cohort update

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/CohortService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/CohortService.kt
@@ -1,10 +1,12 @@
 package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.service
 
 import org.slf4j.LoggerFactory
+import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.OffenceCohort
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.PniScore
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.oasysApi.model.RiskScoreLevel
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.config.AuditorContext
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.entity.ReferralCohortHistoryEntity
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.entity.ReferralEntity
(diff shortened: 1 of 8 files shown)


miroslav25a (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/CohortService.kt
```diff
@@ -19,14 +21,29 @@ class CohortService(
...
+    referralEntity: ReferralEntity,
+    cohort: OffenceCohort,
+    createdBy: String = SecurityContextHolder.getContext().authentication?.name ?: "UNKNOWN_USER",
+  ): ReferralEntity {
+    // Overwrite the username to be written when system is automatically updating this value
+    AuditorContext.set(createdBy)
+    try {
+      val latestCohortHistory =
+        referralCohortHistoryRepository.findTopByReferralIdOrderByCreatedAtDesc(referralEntity.id!!)
+      if (latestCohortHistory?.cohort != cohort) {
```
raby, line 34: The logic on line 34, only saving when the cohort has actually changed in updateCohortForReferral, but CohortServiceTest doesn't cover it. Worth adding two tests to CohortServiceTest?

LLM:
`@Test
fun `saves new history entry when cohort has changed`() {
  val referral = ReferralEntityFactory().produce()
  `when`(referralCohortHistoryRepository.findTopByReferralIdOrderByCreatedAtDesc(referral.id!!))
    .thenReturn(ReferralCohortHistoryFactory().withCohort(OffenceCohort.GENERAL_OFFENCE).produce())

  cohortService.updateCohortForReferral(referral, OffenceCohort.SEXUAL_OFFENCE, "user1")

  verify(referralCohortHistoryRepository).save(any())
}

@Test
fun `does not save when cohort is unchanged`() {
  val referral = ReferralEntityFactory().produce()
  `when`(referralCohortHistoryRepository.findTopByReferralIdOrderByCreatedAtDesc(referral.id!!))
    .thenReturn(ReferralCohortHistoryFactory().withCohort(OffenceCohort.SEXUAL_OFFENCE).produce())

  cohortService.updateCohortForReferral(referral, OffenceCohort.SEXUAL_OFFENCE, "user1")

  verify(referralCohortHistoryRepository, never()).save(any())
}`

Jack-Dempster: there's tests covering this scenario in the controller integration tests. There's quite a bit of duplication across these files

raby: Sure good call

raby (approved):

PR #657, opened 2026-05-19
Sar test and cohort additions

rgforsyth:
adds more data to sar test for completeness.

adds cohort history to sar template


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/model/subjectAccessRequest/SubjectAccessRequestSessionAttendance.kt
```diff
@@ -22,7 +22,7 @@ fun SessionAttendanceEntity.toApi() = SubjectAccessRequestSessionAttendance(
   sessionId = session.id,
   groupMembershipId = groupMembership.id,
   outcomeType = outcomeType.toApi(),
-  legitimateAbsence = legitimateAbsence,
+  legitimateAbsence = legitimateAbsence.toString(),
```
raby, line 25: An AI flagged this null as an issue as null would be "null" in a string. Which is I guess what we intend in this use case :-) so all good

src/main/resources/sar_template.mustache
```diff
@@ -233,30 +253,175 @@
   <p>No Data Held</p>
   {{/attendees.0}}
 
-  <h3>Message History</h3>
-  {{#messageHistories.0}}
+  <h3>Status History</h3>
```
raby: Introducing a full duplicate of Status History, LDC History, Cohort History, Programme Group Memberships, and Sessions Scheduled (lines 256–425) inside the same {{#referrals}} loop, is this intentional? Just checking not an error :)

raby (approved):
Great work

PR #658, opened 2026-05-20
APG-2158 Fix all cohort values being assigned to SEXUAL_OFFENCE 

Jack-Dempster:
Currently all referrals which have a completed OASYs assessment are being assigned to the `SEXUAL_OFFENCE` cohort. This is due to when calculating the cohort the ENUM object is being used as the value itself
```kotlin
RiskScoreLevel.NOT_APPLICABLE.toString()
```
This gives a value = `NOT_APPLICABLE` whereas the value being returned from the Oasys endpoint looks like `Not Applicable` which is what we have called the `type` on the ENUM class. 

The fix is to use the `type` rather than the `.toString()` value when calculating the cohort.

## NOTE

When checking any referrals currently in prod there are some that are `GENERAL_OFFENCE` this is because these referrals are missing the OASYs assessment and therefore the values are null in which we always default to `GENERAL_OFFENCE`


miroslav25a (approved):

raby (commented):
LGTM

PR #659, opened 2026-05-20
APG-2291/prevent cross-region referrals appearing on group allocations waitlist

raby:
**What**
Fixes a data access control bug where the Group Allocations Waitlist tab was displaying referrals from all regions, not just the logged-in user's region. An East Midlands programme team member could see Greater Manchester referrals and PDUs in their waitlist.

**Why**
GetGroupWaitlistItemSpecification had no predicate on regionName in the WAITLIST branch, despite the column existing and being indexed on group_waitlist_item_view. The PDU/reporting team filter dropdown had the same problem — getPdusAndReportingTeams() returned all PDUs from all regions with no constraint.

**How**
- Resolve the authenticated user's region in ProgrammeGroupController via authenticationUtils.getUsername() and pass it down the call chain
- Add regionName = userRegionName predicate to the WAITLIST branch of GetGroupWaitlistItemSpecification
- Add getPdusAndReportingTeamsByRegion(regionName) to ReferralReportingLocationRepository scoped by region, used by getGroupAllocationsFilters
- Retain the existing no-arg getPdusAndReportingTeams() for ReferralCaseListItemService (that filter dropdown has the same issue but is out of scope for this bug fix — see note below)

**Tests**
Two new integration tests added to ProgrammeGroupControllerIntegrationTest:
-WAITLIST only returns referrals from the user's own region — seeds referrals in two regions both with Awaiting allocation status, asserts only the user's region are returned
-WAITLIST filter dropdown only shows PDUs from the user's own region — asserts cross-region PDUs do not appear in filters.locationFilterValues

Existing tests updated to set regionName = "WIREMOCKED REGION" in initialiseReferrals() to align with the wiremocked user in beforeEach.

**Notes**
getCaseListFilterData() in ReferralCaseListItemService has the same cross-region PDU dropdown issue on the main caselist. Left intentionally out of scope to keep this fix focused — tracked separately.


pfcairns (approved):
Good work 👍

PR #662, opened 2026-05-21
APG-2289 Restrict case list PDU filters to user's region

pfcairns:
* Update case list filters to return only PDUs for the authenticated user's region. 
* Adjust accompanying tests to reflect this change.
* Remove deprecated "/bff/caselist/filters" endpoint
* Remove unused ReferralReportingLocationRepository function


miroslav25a (approved):

raby (approved):
Looks great!

PR #663, opened 2026-05-21
Reporting endpoints: Facilitator Consistency & Dosage

thomaswilsonxyz:
**WHAT**

This commit introduces two reporting endpoints:

* `facilitator-consistency`, which provides information about the facilitators for each session 
* `dosage`, which breaks down the number of sessions completed by an individual Person on Probation

**WHY THE BIG PR?**

Great question.  I stacked my PRs (i.e. `a -> b -> main`), and had auto-merge enabled on my `a -> b` PR.  The requirement for a review/approval on a PR to merge is only present on PRs that target `main`, so now I am (accidentally) left with a single (large) feature branch.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReportingController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReportingController.kt
@@ -36,6 +36,7 @@ class ReportingController(
 
     private const val CSV_MEDIA_TYPE = "text/csv"
     private const val GROUP_SIZE_FILE_NAME_SUFFIX = "manage-and-deliver-group-size.csv"
+    private const val FACILITATOR_CONTINUITY_FILE_NAME_SUFFIX = "facilitator-continuity.csv"
     private const val DOSAGE_FILE_NAME_SUFFIX = "dosage.csv"
   }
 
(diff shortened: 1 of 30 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReportingController.kt
```diff
@@ -95,4 +98,160 @@ class ReportingController(
...
+        content = [Content(schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "401",
+        description = "Unauthorized",
+        content = [Content(schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  @PreAuthorize("hasAnyRole('ROLE_ACCREDITED_PROGRAMMES_MANAGE_AND_DELIVER_API__ACPMAD_UI_WR')")
```
pfcairns, line 223: Didn't we set up a separate role for the reporting endpoints, or did I imagine this?

thomaswilsonxyz: You didn't imagine it - that's controlled in the UI :) this is just saying that the UI client can access this endpoint, it's up to the UI to control the (user's) roles for accessing endpoints there

pfcairns: Yes, sorry that makes perfect sense.

pfcairns (approved):
The Controller looks fine, as do the Repository implementations. The monster query and use of JdbcTemplate in the private `ReportingService` method  `getFacilitatorContinuityReportRows` isn't the easiest to understand (or maintain) without a JIRA ticket or some requirements for context, so will have to trust that the integration tests are doing their job here.  Further down the line maybe this could be split into constituent repo methods, with the persistence logic moving out of the service, and using JPA rather than the slightly "old school" JDBCTemplate way of doing things.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReportingService.kt
```diff
@@ -3,18 +3,34 @@ package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.ser
...
+  private val sessionAttendanceRepository: SessionAttendanceRepository,
+  private val moduleSessionTemplateRepository: ModuleSessionTemplateRepository,
+  private val namedParameterJdbcTemplate: NamedParameterJdbcTemplate,
+  private val clock: Clock,
 ) {
   companion object {
     private const val GROUP_SIZE_CSV_HEADER =
       "id,code,createdAt,sex,cohort,isLdc,earliestPossibleStartDate,regionName,pduCode,pduName,locationCode,locationName,groupSize,facilitatorStaffCode"
+    private const val FACILITATOR_CONTINUITY_CSV_HEADER =
+      "code,sessionNumber,sessionName,sessionType,isCatchUp,attendeeCount,facilitatorStaffCodes,region_name,delivery_location_name,probation_delivery_unit_name,sessionStartTime,sessionCreatedAt"
```
raby, line 33: I had an LLM comment to me regarding inconsistent use of camel case here and _ delimiter elsewhere. Feel free to ignore if not useful.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/ReferralRepository.kt
```diff
@@ -22,4 +23,35 @@ interface ReferralRepository : JpaRepository<ReferralEntity, UUID> {
...
+             r.event_id AS licReqNo,
+             r.crn AS crn,
+             COALESCE(rrl.region_name, 'UNKNOWN_REGION_NAME') AS regionName
+      FROM referral r
+      LEFT JOIN referral_reporting_location rrl ON r.id = rrl.referral_id
+      LEFT JOIN referral_status_history rsh ON r.id = rsh.referral_id
+      LEFT JOIN referral_status_description rsd ON rsd.id = rsh.referral_status_description_id
+      WHERE r.intervention_name = 'Building Choices'
+        AND (
+          (CAST(:referralsCreatedSince AS DATE) IS NOT NULL AND r.created_at::date > CAST(:referralsCreatedSince AS DATE))
```
raby, line 39: The parameter is named referralsCreatedSince and the controller documents it as "Only include referrals created after this date". The semantic of "since" is typically inclusive (>=). The controller doc says "after" but the query param is named "since". There's an ambiguity here that could lead to off-by-one errors (a referral created exactly on the boundary date would be excluded).

If "since" means inclusive, change > to >= in the query and update the controller OpenAPI description to match. Or rename the parameter to referralsCreatedAfter consistently?

A small thing!

raby (commented):
Really good work - just a couple of small comments

PR #665, opened 2026-05-21
Apg 2251/track user actions using the telemetry client in app insights and fix the lazy loading exception 

miroslav25a:


Jack-Dempster (approved):
👍 Thanks for sorting that issue so quickly

PR #666, opened 2026-05-21
Refactor `getFirstUserRegionDescription` into `UserService`

pfcairns:
Refactors the `getFirstUserRegionDescription` method, moving it into the `UserService` class to improve code organization and maintainability. All references to the method have been updated accordingly.
This will give some logging also to see which region is being used for the case list endpoint.


raby (approved):
Lack of a UserService test file but I feel thats another PRs scope so passing , good logging work!

PR #670, opened 2026-05-22
Refine `GetProgrammeGroupsSpecification` and add soft-delete support

pfcairns:
- **Summary of Changes**
  - Updated `GetProgrammeGroupsSpecification` to improve membership filtering logic.
  - Added support for handling soft-deleted memberships.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/specification/GetProgrammeGroupsSpecification.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/specification/GetProgrammeGroupsSpecification.kt
@@ -25,7 +25,7 @@ import java.time.LocalDate
 import java.time.LocalDateTime
 import kotlin.jvm.java
 
-fun hasAtLeastOneMembership(
+fun hasAtLeastOneActiveMembership(
   query: CriteriaQuery<*>,
   cb: CriteriaBuilder,
   root: Root<ProgrammeGroupEntity>,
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/specification/GetProgrammeGroupsSpecification.kt
```diff
@@ -211,16 +211,32 @@ fun getProgrammeGroupsByRegionTabSpecification(
       cb.or(
         notStartedOrInProgressDateSpec,
         cb.greaterThan(incompleteMembershipCountSubquery, 0L),
-        cb.not(hasAtLeastOneMembership(query, cb, root)),
+        cb.not(hasAtLeastOneActiveMembership(query, cb, root)),
```
raby, line 214: A group with deleted memberships where all deleted members also happen to meet the `isMembershipComplete` criteria (programme complete status + PPR attended + all sessions have attendance). In that case `incompleteMembershipCountSubquery == 0`, COMPLETE matches, and NOT_STARTED_OR_IN_PROGRESS also matches via `cb.not(hasAtLeastOneActiveMembership)`.

My understanding might be wrong, but I wondered if a group with only deleted memberships could satisfy both branches:
cb.not(hasAtLeastOneActiveMembership(...)) 
hasAtLeastOneMembershipIncludingDeleted(...) 

If that’s expected behaviour, all good 👍
If not, would we want the tabs to be mutually exclusive?

Maybe I am overthinking.

pfcairns: Yes, we do want them to be mutually exclusive, although I'm less concerned about the NOT_STARTED_OR_IN_PROGRESS count at present, as the COMPLETE tab is still not displaying any groups, despite their being some. I'll address your comment in another PR if that's OK?

raby: Yes sure makes sense a small separate PR

raby (approved):

PR #671, opened 2026-05-26
add personName for session notes as an individual field

Alison-Weatherston:
This field is required for the page titles

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/model/sessionNotes/SessionNotes.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/model/sessionNotes/SessionNotes.kt
@@ -11,16 +11,16 @@ import java.util.UUID
 @Schema(description = "Details of the notes for a session")
 data class SessionNotes(
   @Schema(
-    description = "The person on probation's name",
-    example = "Alex River",
-  )
-  val personName: String,
-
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/model/sessionNotes/SessionNotes.kt
```diff
@@ -10,6 +10,12 @@ import java.util.UUID
 
 @Schema(description = "Details of the notes for a session")
 data class SessionNotes(
+  @Schema(
+    description = "The person on probation's name",
+    example = "Alex River",
+  )
+  val personName: String,
```
raby: Nit / non-blocking 🙌

`personName` is currently placed as the first field in the data class, which means it'll be the first key in the JSON response. Totally fine functionally (the UI deserialises by key name not position), but for consistency with the rest of the codebase it might read more naturally after `pageTitle` since that's what leads the page context:

```kotlin
data class SessionNotes(
  val pageTitle: String,
  val personName: String,  // moved here
  val moduleName: String,
  ...
)

Alison-Weatherston: Thanks, @raby. personName moved to the top

raby (approved):

PR #673, opened 2026-05-26
fix(APG-2307): gracefully handle OASys PNI failures to prevent 500s

raby:
fix(APG-2307): gracefully handle OASys PNI failures to prevent 500s

## What

Catches OASys PNI exceptions (503 Service Unavailable) inside the `async` coroutine block so that the referral details page loads with safe defaults instead of crashing with a 500 error.

## Why

App Insights data (7 days, 26 May 2026) shows:
- **199,644 × 404** from OASys (no assessment exists for most CRNs)
- **8,201 × 503** from OASys (periodic unavailability)
- Only **11 actual timeouts** — the reported "timeout" bug is really an unhandled exception problem

The 503s were propagating as 500s to the UI, preventing staff from viewing any referral when OASys was temporarily down.

## How

- Move try-catch **inside** the `async(Dispatchers.IO)` block (not around `.await()`) — this is required because `coroutineScope` propagates child exceptions immediately, cancelling all sibling coroutines before `await()` is reached
- On failure: log warning, return `null`, use defaults (General Offence cohort, No LDC)
- Page always loads — nDelius personal details / sentence are unaffected

## Test coverage

- New integration test: `retrieve referralDetails gracefully when OASys PNI returns 503`
- New WireMock stub: `stubServiceUnavailablePniResponse`

## Relationship to long-term fix

This is a **strict subset** of `APG-2307/pni-fix-500-error` (the long-term branch). That branch adds daily-freshness guard (skip OASys if already enriched today) on top of this same try-catch. Merge this first for immediate production relief; the long-term branch merges cleanly on top.

## Impact

| Before | After |
|--------|-------|
| 8,200 page crashes/week from OASys 503 | **0** — page loads with defaults |

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralService.kt
@@ -88,7 +88,16 @@ class ReferralService(
     val referral = referralRepository.findByIdWithMemberships(referralId) ?: return@coroutineScope null
 
     val pniDeferred = async(Dispatchers.IO) {
-      pniService.getPniResponse(referral.crn)
+      try {
+        pniService.getPniResponse(referral.crn)
+      } catch (e: Exception) {
+        // APG-2307: Gracefully handle OASys PNI failures (503s were crashing the page with 500).
+        // Catching broad Exception here — PniService only throws ClientResult.Failure.toException() which wraps
+        // HTTP errors, so this is safe. Could narrow to WebClientResponseException if we want to surface
+        // unexpected bugs (e.g., NPE) rather than swallowing them. Follow-up refinement candidate.
+        log.warn("Failed to fetch PNI for referral $referralId (CRN: ${referral.crn}): ${e.message}")
+        null
+      }
     }
 
     val personalDetailsDeferred = async(Dispatchers.IO) {
@@ -127,6 +136,7 @@ class ReferralService(
     val latestReferralCohort =
       referralCohortHistoryRepository.findTopByReferralIdOrderByCreatedAtDesc(referralId)?.cohort
     val allocatedGroup = programmeGroupMembershipService.getCurrentlyAllocatedGroup(referral)
+
     ReferralDetails.toModel(
       referral,
       personalDetails,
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralService.kt
```diff
@@ -88,7 +88,16 @@ class ReferralService(
     val referral = referralRepository.findByIdWithMemberships(referralId) ?: return@coroutineScope null
 
     val pniDeferred = async(Dispatchers.IO) {
-      pniService.getPniResponse(referral.crn)
+      try {
```
Jack-Dempster: doing this and setting it to null will actually overwrite the referral cohort to GENERAL_OFFENCE. We probably need to change it further down so we don't do this

raby: Thank you good spot

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralService.kt
```diff
@@ -88,7 +88,16 @@ class ReferralService(
     val referral = referralRepository.findByIdWithMemberships(referralId) ?: return@coroutineScope null
 
     val pniDeferred = async(Dispatchers.IO) {
-      pniService.getPniResponse(referral.crn)
```
pfcairns, line 91: As an alernative solution and additional fix, the `PniService.getPniResponse(..)`  function could be modified to also return a null in the event of any downstream failure, rather than just the 404 status which is the current case. For this to work, all clients should call the `getPniCalculation` method rather than `getPniResponse` which should really be private in the service class. Happy to discuss if the above doesn't make any sense.

pfcairns (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralService.kt
raby: I have pushed up a commit to hopefully address the default override cohort behaviour with a test, cheers.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralService.kt
raby: I have pushed up a commit to hopefully address this pattern , which is a good suggestion

pfcairns (approved):
Thanks for making the changes.

PR #687, opened 2026-05-29
use helper function to only show last name of facilitators

rgforsyth:
Use SAR helper function to only show surnames of staff members, by using their id.


raby (approved):
Looks great

PR #688, opened 2026-05-29
Add `BankHolidayEntity` and `BankHolidayRepository` for bank holiday management

pfcairns:
### Summary
- Created a new `bank_holiday` table and populated it with data from a JSON file (retrieved from https://www.gov.uk/bank-holidays.json)
- Implemented `BankHolidayEntity` and `BankHolidayRepository` to manage bank holiday data.
- Updated `ScheduleService` to fetch bank holidays from the database while keeping the API client intact, so we can use it further down the line from an admin API endpoint or similar to refresh the DB data periodically

### Changes
- Added Flyway migration script `V103__create_bank_holiday_table.sql` to create the `bank_holiday` table and insert 83 bank holiday records for England and Wales. This data set takes us up to Boxing Day, 2028 which should be ample for private beta.
- Created `BankHolidayEntity.kt` with fields for `id`, `title`, `holidayDate`, `notes`, and `bunting`.
- Created `BankHolidayRepository.kt` as a standard `JpaRepository`.
- Modified `ScheduleService.kt` to inject `BankHolidayRepository` and use it in `englandAndWalesHolidayDates()`.
- Renamed the previous API-based implementation to `englandAndWalesHolidayDatesFromApi()` in `ScheduleService.kt` to keep it available for future use.


raby (approved):
Really good clean work

PR #690, opened 2026-05-29
APG-2311 Add validation to block rescheduling of past sessions

pfcairns:
---

### Summary

This pull request adds validation to prevent the rescheduling of past sessions.
Updates include the ability to reschedule the start and end times of a session as long as it is not in the past.
It also includes updates and enhancements to the session rescheduling tests to ensure full coverage for this new functionality.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
@@ -120,7 +120,7 @@ class SessionService(
     // validate that the reschedule request is valid - session duration cannot be changed for past sessions
     if (session.startsAt.isBefore(LocalDateTime.now())) {
       log.warn("Invalid reschedule request received for past session with id: $sessionId. Requested session start date must be in the future")
-      return EditSessionDateAndTimeResponse("The session session duration cannot be longer than originally scheduled. Change the start or end time.")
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
```diff
@@ -111,14 +111,21 @@ class SessionService(
...
 
     val originalSessionStartsAt = session.startsAt
 
-    // update the start time of the requested session
-    session.startsAt = requestedStartsAt
-    session.endsAt = session.endsAt.plus(startOffset)
+    // validate that the reschedule request is valid - session duration cannot be changed for past sessions
+    if (session.startsAt.isBefore(LocalDateTime.now())) {
+      log.warn("Invalid reschedule request received for past session with id: $sessionId. Requested session start date must be in the future")
+      return EditSessionDateAndTimeResponse("The session session duration cannot be longer than originally scheduled. Change the start or end time.")
```
Jack-Dempster: will this not return a 200? Should we not throw a 400 here instead so it the ui knows to display something?

Jack-Dempster (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
pfcairns: Good point. Will address.

Jack-Dempster (approved):
👍

PR #701, opened 2026-06-02
Update session rescheduling logic to prevent past session changes

pfcairns:
Update the session rescheduling logic to ensure that duration changes are prevented for sessions that have already occurred, but allowing past sessions to be rescheduled where the duration remains the same.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/SessionControllerIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/SessionControllerIntegrationTest.kt
@@ -309,7 +309,7 @@ class SessionControllerIntegrationTest : IntegrationTestBase() {
   @Test
   fun `should return HTTP 400 code when attempting to amend the duration of a session in the past `() {
     // Given
-    val programmeTemplate = accreditedProgrammeTemplateRepository.getBuildingChoicesTemplate()
+    val programmeTemplate = testDataGenerator.createAccreditedProgrammeTemplate("Test Programme")
     val module = testDataGenerator.createModule(programmeTemplate, "Test Module", 1)
     val sessionTemplate = testDataGenerator.createModuleSessionTemplate(
       ModuleSessionTemplateEntity(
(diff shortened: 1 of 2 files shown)


miroslav25a (dismissed):

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/SessionControllerIntegrationTest.kt
```diff
@@ -356,6 +356,56 @@ class SessionControllerIntegrationTest : IntegrationTestBase() {
...
+    // When
+    val response = performRequestAndExpectStatusWithBody(
+      httpMethod = HttpMethod.PUT,
+      uri = "/session/${session.id}/reschedule",
+      body = rescheduleRequest,
+      returnType = object : ParameterizedTypeReference<EditSessionDateAndTimeResponse>() {},
+      expectedResponseStatus = HttpStatus.OK.value(),
+    )
+
+    // Then
```
raby, line 405: The test asserts the response message is correct, but doesn't assert that session.endsAt was actually updated to preserve the original 1-hour duration. Since this is the core behaviour of the ticket (allow rescheduling past sessions without changing duration), worth asserting the actual persisted times:

LLM
```
> // Verify the session was rescheduled and duration was preserved (original was 1hr: 13:30–14:30)
> val updatedSession = sessionRepository.findById(session.id).get()
> val expectedStart = LocalDateTime.of(LocalDate.now().plusDays(2), LocalTime.of(10, 0))
> val expectedEnd = expectedStart.plusHours(1) // original duration preserved
> assertThat(updatedSession.startsAt).isEqualTo(expectedStart)
> assertThat(updatedSession.endsAt).isEqualTo(expectedEnd)
```

pfcairns: Thanks Raby, this behaviour is tested in other integration tests, hence it's omission.

pfcairns: Have addressed this now in any case.

raby (approved):

PR #705, opened 2026-06-03
no ticket/2026 06 03 mvc timeout 3m

thomaswilsonxyz:
- **tweak: Set the timeout on WebClient from 20s -> 60s**
- **config: set spring.mvc.request-timeout to 3 minutes**


Jack-Dempster (approved):
👍

PR #706, opened 2026-06-03
fix: Make onboarding endpoint fire-and-forget to prevent AsyncRequestTimeoutException

raby:
Title: fix: Make onboarding endpoint fire-and-forget to prevent AsyncRequestTimeoutException
Description:

**Problem**

The POST /onboarding/referrals endpoint was returning 500 errors in production due to AsyncRequestTimeoutException.
When processing 5-6 referrals sequentially, each nDelius call takes ~10s (especially for CRNs with eventNumber: 0). Total processing time of ~50-60s exceeds Spring's default 30s async request timeout, causing the coroutine to be cancelled mid-flight and returning a 500 to the caller.

**Production evidence:**
ERROR AdminService : [Referral 5310bf3a...]...failure: MonoCoroutine was cancelled)
ERROR ApiExceptionHandler : Unexpected error
org.springframework.web.context.request.async.AsyncRequestTimeoutException

**Fix**
OnboardingController: Changed to fire-and-forget pattern using CoroutineScope(dispatcher).launch — matches the existing pattern in AdminController. The endpoint now returns 200 immediately and processes referrals in the background.
application.yml: Added spring.mvc.async.request-timeout: 120000 as a safety net for any other suspend endpoints.

**API Contract**
No breaking changes. Response shape remains {successIds: [], notFoundIds: [], failureIds: []} with status 200. The lists will now always be empty (since processing is async), but the shape is unchanged.

**Notes**
The WebClient timeout increase (20s → 60s) in the previous commit does not fix this issue — individual nDelius calls complete fine (~10s), it's the aggregate sequential processing time that exceeds Spring's async request timeout.
The AdminController already uses this exact fire-and-forget pattern and has no issues.
Background processing results are logged for observability.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/OnboardingController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/OnboardingController.kt
@@ -4,6 +4,10 @@ import com.fasterxml.jackson.annotation.JsonProperty
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.Parameter
 import io.swagger.v3.oas.annotations.media.Schema
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import org.slf4j.LoggerFactory
 import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/OnboardingController.kt
```diff
@@ -66,25 +71,28 @@ class OnboardingController(
...
+          result.successIds.size,
+          result.notFoundIds.size,
+          result.failureIds.size,
+        )
+      } catch (e: Exception) {
+        log.error("Error during background processing of personal details for referrals: ${request.referralIds}", e)
+      }
     }
+
+    return ResponseEntity.ok(
```
Jack-Dempster: I think 202 is .accepted?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/OnboardingController.kt
```diff
@@ -66,25 +71,28 @@ class OnboardingController(
...
-      return ResponseEntity.ok(
-        FetchPersonalDetailsResponse(
-          successIds = emptyList(),
-          notFoundIds = emptyList(),
-          failureIds = request.referralIds.map(UUID::toString),
-        ),
-      )
+    log.info("Accepted {} referrals for background processing: {}", request.referralIds.size, request.referralIds)
+
+    CoroutineScope(dispatcher).launch {
```
Jack-Dempster, line 75: since we are just firing and forget now. I'd suggest just deleting the `FetchPersonalDetailsResponse` data class and returning just a string or whatever we would normally return as empty response

raby: Yes good spot. replaced it with `FetchPersonalDetailsAcceptedResponse` which just has message and referralCount. Cheers

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/OnboardingController.kt
raby: I have made this change.

Jack-Dempster (approved):

PR #711, opened 2026-06-04
perf: cache getUserRegions() with Caffeine (5 min TTL) 

raby:
PR: Cache getUserRegions() + fix empty region crash

**What**
Added @Cacheable to getUserRegions() in UserService.kt — caches user region lookups by username, avoiding repeated nDelius calls for the same user within a session. Uses the user-regions cache with Caffeine TTL.

Added Caffeine cache dependency to build.gradle.kts with TTL configuration in application.yml.

Fixed IndexOutOfBoundsException in ProgrammeGroupService.kt — when getUserRegions() returns an empty list (e.g. nDelius times out and returns gracefully), the destructuring val (userRegion) = ... crashed with Empty list doesn't contain element at index 0. Now throws a proper NotFoundException instead of an unhandled 500.

**Why**
Production logs show ReadTimeoutException on GET /user/{username}/teams (nDelius) causing getUserRegions() to return emptyList(). The downstream destructuring in ProgrammeGroupService.getGroupWaitlistDataByCriteria() then crashes, surfacing as a 500 to the UI ("Socket timeout" from the UI's perspective).

Caching eliminates repeated calls for the same user, and the empty-list guard prevents the crash when nDelius is transiently unavailable.

**Evidence from prod logs**
```
io.netty.handler.timeout.ReadTimeoutException
→ UserService: Failed to fetch teams for user GLM19A
→ IndexOutOfBoundsException: Empty list doesn't contain element at index 0
  at ProgrammeGroupService.getGroupWaitlistDataByCriteria(ProgrammeGroupService.kt:354)
```
**Testing**
Existing tests pass
Cache only stores non-empty results (unless = "#result.isEmpty()")
On failure, returns 404 with clear message rather than 500

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/config/CacheConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/config/CacheConfiguration.kt
@@ -1,0 +1,39 @@
+package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.config
+
+import com.github.benmanes.caffeine.cache.Caffeine
+import org.springframework.cache.CacheManager
+import org.springframework.cache.caffeine.CaffeineCacheManager
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import java.util.concurrent.TimeUnit
+
+@Configuration
+class CacheConfiguration {
+
+  @Bean
+  fun cacheManager(): CacheManager {
+    val cacheManager = CaffeineCacheManager()
+    cacheManager.registerCustomCache(
+      "user-regions",
+      Caffeine.newBuilder()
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/config/CacheConfiguration.kt
Jack-Dempster, line 1: is there a reason we are using a new dependency rather than just using spring caching?

raby: The alternative without Caffeine would be Spring's built-in `ConcurrentMapCache` — but that has no TTL support. 

The @Bean approach is needed here specifically because our three caches have different TTLs (5min, 24h, 24h) and sizes, which can't be expressed with a single spec string. An alternative is to use `spring.cache.caffeine.spec ` in application.yml but applies to all caches

PR #716, opened 2026-06-08
Apg 2329/fix bug when username without region

miroslav25a:


pfcairns (dismissed):
LGTM

raby (dismissed):
✅ Looks good overall. All sites guarded, nice test coverage.

Minor thought: For the BFF dropdown endpoints (/bff/pdus-for-user-region, /bff/region/members, schedule-session facilitators)  throwing NotFoundException means the UI gets a 404 and shows an error page.

An alternative though could be returning an empty list `(ResponseEntity.ok(emptyList())) `so the page still loads with empty dropdowns when nDelius is transiently unavailable. The user can still navigate and retry.

Not blocking, happy either way, just flagging the UX trade-off for consideration if not been thought of yet by the wider team.

DanEgan1 (dismissed):

raby (approved):
looks great

PR #717, opened 2026-06-08
APG-2292 : set uniquness of facilitator list to be personCode -> personName

Jack-Dempster:
When we get a list of facilitators we were originally ensuring these are unique by team code and person code. However this is returning duplicate entries for people in multiple teams. This PR fixes this as Lauren confirmed that when selecting the facilitor/TM the team doesn't matter


raby (approved):
LGTM

PR #721, opened 2026-06-09
Apg 2332/graceful ndelius appointment failure

raby:
APG-2332: Handle nDelius appointment failure gracefully during group allocation

Wrap nDelius appointment creation in try-catch so group allocation
succeeds even when nDelius returns 400/503. Log error and emit
telemetry for failed appointments to enable manual sync.

Adds unit test verifying allocation completes when nDelius is unavailable.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupMembershipService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupMembershipService.kt
@@ -90,10 +90,28 @@ class ProgrammeGroupMembershipService(
       attendeeEntity
     }
 
-    // Create appointments in NDelius for each session object
-    scheduleService.createNdeliusAppointmentsForSessions(newAttendees)
-
+    // Save the referral first to persist group membership and status change
     val savedReferral = referralRepository.save(referral)
+
+    // Create appointments in NDelius — non-blocking so allocation still succeeds if nDelius is unavailable
+    try {
+      scheduleService.createNdeliusAppointmentsForSessions(newAttendees)
+    } catch (ex: Exception) {
+      log.error(
+        "Failed to create nDelius appointments for referral $referralId in group $groupId. " +
+          "Group allocation succeeded but appointments need manual sync. Error: ${ex.message}",
+        ex,
+      )
+      telemetryClient.logToAppInsights(
+        "Referral.allocate-to-group.ndelius-appointment-failure",
+        mapOf(
+          "activityType" to ASSIGN_REFERRAL_TO_GROUP.name,
+          "referralId" to referralId.toString(),
+          "groupId" to groupId.toString(),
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupMembershipService.kt
```diff
@@ -90,11 +90,29 @@ class ProgrammeGroupMembershipService(
...
-    scheduleService.createNdeliusAppointmentsForSessions(newAttendees)
-
+    // Save the referral first to persist group membership and status change
     val savedReferral = referralRepository.save(referral)
 
+    // Create appointments in NDelius — non-blocking so allocation still succeeds if nDelius is unavailable
+    try {
+      scheduleService.createNdeliusAppointmentsForSessions(newAttendees)
+    } catch (ex: Exception) {
+      log.error(
```
Jack-Dempster, line 100: we definitely don't want to do this. If we can't create sessions in delius then we will have problems down the line without any functionality to try and re create them in the future. 

I don't think this is the correct behaviour here and we should fail if we can't create appointments

raby: Thanks Jack. Yes we don't want to lose synchronisation between us and ndelius with just the logs and humans picking up logs knowing about it. I was thinking of a pending status on an nDelusAppointment which then an every 5 min job executes and tries again. This works for some errors. Some are 404 though due to some data we are sending is incorrect and investigating that today, even if the 5min job executes with 404 every time keeping in pending state then requires an admin portal for stuck pendings- I have been realising this is a bigger job then logging an error. I thought logging an error may give some visual improvement to users short term if an urgent issue but yes could create more problems then solves!

Jack-Dempster: Even if we had a retry job there is no guarantee it would ever succeed and the user could be adding attendance and notes to these sessions which we would never be able to add to delius without manual intervention. If delius service is unavailable they wouldn't even be able to login to our service in the first place so any errors here feel like genuine errors that we should surface to the user at the time of error.

raby: You're absolutely right Jack, I'm closing this PR. The login to nDelius is the clincher: if the user can reach the allocation screen, nDelius is up, so any failure here is a genuine data issue that the user needs to know about immediately. Swallowing it would just create a worse problem downstream with attendance/notes against non-existent appointments, yes. The enhanced logging PR (already merged) will give me the response bodies to root-cause the 400s. This is the real fix, preventing the bad requests rather than papering over them. I will create a separate PR/branch when a fix for the 404's becomes apparent in the investigation. Cheers.

PR #727, opened 2026-06-11
Update the tag colours returned for a status

niaevans9:
Update the tag colours returned for a status and return the colour for the caselist
`<div class="tableView-content-wrap" data-prosemirror-initial-todom-render="true" data-prosemirror-content-type="node" data-prosemirror-node-name="table" data-prosemirror-node-block="true" data-pm-slice="1 1 []"><div data-testid="table-alignment-container" style="display: flex;justify-content: center"><div class="pm-table-resizer-container" style="--ak-editor-table-gutter-padding: calc(var(--ak-editor--large-gutter-padding) * 2); --ak-editor-table-width: calc(100cqw - calc(var(--ak-editor--large-gutter-padding) * 2)); width: var(--ak-editor-table-width);"><div class="resizer-item display-handle" style="--ak-editor-table-max-width: 1800px; --ak-editor-table-min-width: 97px; box-sizing: border-box; max-width: 100%; min-width: var(--ak-editor-table-min-width); position: relative; user-select: auto; width: auto;"><span class="resizer-hover-zone"><div class="pm-table-container" data-number-column="false" data-layout="center" data-testid="table-container"><div class="pm-table-sticky-sentinel-top" data-testid="sticky-sentinel-top"></div><div class="pm-table-row-controls-wrapper"><div></div></div><div class="pm-table-wrapper">
Status | Tag Colour
-- | --
Programme complete | govuk-tag--grey
On programme | govuk-tag--green
Scheduled | govuk-tag--blue
Awaiting allocation | govuk-tag--teal
Awaiting assessment | govuk-tag--purple
Withdrawn | govuk-tag--red
Breach | govuk-tag--yellow
Recall | govuk-tag--yellow
Return to court | govuk-tag--yellow
Deferred | govuk-tag--orange
Deprioritised | govuk-tag--orange
Suitable but not ready | govuk-tag--orange

</div><div class="pm-table-sticky-sentinel-bottom" data-testid="sticky-sentinel-bottom"></div></div></span></div></div></div></div>`


raby (commented):
Hey! Great work getting the status colours on the caselist, the migration and test updates all look solid.

I've got one suggestion on the approach for getting `label_colour` into the entity. Right now the PR adds a `@ManyToOne` relationship that joins the status text column back to `referral_status_description.description_text`. This works, but it's a bit fragile because:

- It joins on a text string rather than a primary key, and relies on that text always being unique (there's no DB constraint enforcing that)
- It pulls in the full` ReferralStatusDescriptionEntity `(with `isClosed`, `createdAt`, `deletedAt` etc.) when we only need the one string field

The good news is we already solved this exact problem on the group waitlist view in R__create_group_wait_list_view.sql:

`rsd.description_text  as status,
rsd.label_colour      as status_colour,`

It selects the colour directly in the view, so the entity can map it as a plain column. We can do the same here?

What the fix could look like:

In R__create_referral_caselist_item_view.sql, add the colour to the SELECT as the join to `rsd` exists there:

```
rsd.description_text                                   as status,
rsd.label_colour                                       as status_label_colour,
```

In `ReferralCaseListItemViewEntity`, replace the `@ManyToOne` with a simple column and remove `insertable = false`, `updatable = false` from status:

`@NotNull
@Column(name = "status")
var status: String,

@NotNull
@Column(name = "status_label_colour")
var statusLabelColour: String,`

The toApi() mapper:

`statusLabelColour = statusLabelColour,`

The reason this could be preferable is that materialised views are designed to pre-compute and flatten data, so the optimal place to resolve the colour is in the view's SQL, not at query time via a JPA relationship. It keeps the entity simple, avoids the non-PK join, and matches what we already do elsewhere in the codebase.

I learnt a lot about the codebase looking into this, thank you

raby (approved):
Great work

PR #729, opened 2026-06-12
Apg 2332/pre validate ndelius sentence data before allocation

raby:
APG-2332: Pre-validate nDelius requirement/licence condition before group allocation

**What**
Adds a pre-validation step when allocating a referral to a programme group. Before mutating any state, we now call nDelius to verify that the referral's linked requirement or licence condition still exists.

**Why**
Previously, if a referral's sentence data was stale (e.g. after a transfer or sentence termination), the allocation would partially succeed but then fail with a cryptic 400 error when attempting to create appointments in nDelius. This made it difficult for users to understand what went wrong and left the system in a partially mutated state.
By validating upfront, we fail fast with a clear, user-friendly error message before any state changes occur.

**Changes**
`ProgrammeGroupMembershipService` — Added ``validateReferralSentenceDataExistsInNDelius()` which calls the appropriate nDelius endpoint (`getRequirementManagerDetails` or `getLicenceConditionManagerDetails`) based on the referral's sourcedFrom type. Throws a `BusinessException` with a descriptive message on failure.

`NDeliusCaseRequirementOrLicenceConditionResponseFactory` — New test factory for building nDelius response objects.

`ProgrammeGroupMembershipServiceTest` — Added tests covering the validation logic (success, 404/status failures, network failures, null eventId/sourcedFrom).

`TestReferralHelper` — Extracted common test setup into a reusable helper method.

**Notes**
The extra GET call on every allocation is intentional — it's a lightweight read that prevents expensive failure recovery later.
Internal IDs are not exposed in user-facing error messages.
Success log demoted to debug level to reduce noise.


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupMembershipService.kt
```diff
@@ -206,4 +215,93 @@ class ProgrammeGroupMembershipService(
...
+    when (result) {
+      is ClientResult.Success -> {
+        log.debug("nDelius validation passed for referral ${referral.id}: ${sourcedFrom.name} $eventId exists")
+      }
+
+      is ClientResult.Failure.StatusCode -> {
+        log.error(
+          "nDelius validation failed for referral ${referral.id}: $sourceType with ID $eventId " +
+            "does not exist in nDelius for CRN ${referral.crn} (status: ${result.status})",
+        )
```
pfcairns, line 266: You could additionally log the exception here.

pfcairns (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupMembershipService.kt
raby: Great spot enables more info for debugging! I have included this change, if can re-review cheers

pfcairns (approved):
Thanks for doing this.

PR #731, opened 2026-06-15
Apg 2349/delivery location alpha

Alison-Weatherston:
Return delivery locations in alphabetical order.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
@@ -703,7 +703,7 @@ class ProgrammeGroupService(
       null
     } else {
       programmeGroupRepository.findDistinctDeliveryLocationNames(firstUserRegionDescription, pdu)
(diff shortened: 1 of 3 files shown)


raby (dismissed):
Good

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
```diff
@@ -703,6 +703,7 @@ class ProgrammeGroupService(
       null
     } else {
       programmeGroupRepository.findDistinctDeliveryLocationNames(firstUserRegionDescription, pdu)
+        .sortedWith(String.CASE_INSENSITIVE_ORDER)
```
Jack-Dempster, line 706: I would probably do the sorting at the repository level as this may be called in a few places

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
Jack-Dempster, line 700: we should sort this alphabetically too

Jack-Dempster (approved):
Thanks for making those changes

PR #733, opened 2026-06-15
added test that causes deadlock errors

DanEgan1:


src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReferralControllerIntegrationTest.kt
```diff
@@ -1118,6 +1118,145 @@ class ReferralControllerIntegrationTest(@Autowired private val programmeGroupMem
     }
   }
 
+  @Nested
+  @DisplayName("Multi-Endpoint Concurrent Requests avoid deadlock")
+  inner class MultiEndpointConcurrentRequests {
+    @Test
+    fun `should handle concurrent requests to multiple different endpoints without deadlock`() {
+      // Test may have issues when run as part of full build due to large number of calls.
```
DanEgan1, line 1126: Issue I noticed whilst developing Initially used a high number of calls to cause deadlock. I was abole to reduce the number of calls whilst still deadlocking so it works fine in build now but thought its a good idea to leave a warning just in case

pfcairns (approved):
Good work. We may wish to disable the tests if they are long running as you state. It'll be worth checking the logs in dev and pre-prod for instances of the error in the RAISE WARNING command before shipping to Prod.

raby (approved):
Looks good!

PR #735, opened 2026-06-16
fix: use expectedEndDate for LICENCE_CONDITION sentence end date

raby:
Fix: Use expectedEndDate for LICENCE_CONDITION sentence end date

**What**
Previously, LICENCE_CONDITION referrals incorrectly used licenceExpiryDate as the sentence end date, while REQUIREMENT referrals correctly used expectedEndDate. Both referral types should use expectedEndDate.

**Why**
licenceExpiryDate serves a different purpose — it is used to calculate postSentenceSupervisionStartDate (licence expiry date + 1 day). It should not be used as the sentence end date for licence condition referrals.

**Changes**
SentenceService.kt — Updated getSentenceEndDate() so both REQUIREMENT and LICENCE_CONDITION return expectedEndDate
NDeliusSentenceResponseFactory.kt — Simplified factory: expectedEndDate is now always populated regardless of sourcedFrom. licenceExpiryDate remains conditional (only set for LICENCE_CONDITION) as it's used for other calculations
TestReferralHelper.kt — Removed branching logic that set licenceExpiryDate vs expectedEndDate based on sourcedFrom; now always sets expectedEndDate

Integration tests — Updated all test stubs that were using withLicenceExpiryDate() to use withExpectedEndDate() for sentence end date assertions

**Testing**
Updated unit and integration tests to reflect the corrected behaviour
All existing tests pass with the updated logic


Jack-Dempster (approved):
Looks good mate 👍

PR #742, opened 2026-06-17
Update session rescheduling to maintain cadence during shifts

pfcairns:
The pull request addresses the feature update for rescheduling group sessions. It ensures that session cadence is preserved when shifts occur, taking into account bank holidays and the like.


DanEgan1 (approved):

raby (approved):
Looks great!

PR #743, opened 2026-06-18
Skip ndelius appointments creation on referral allocation for past sessions

pfcairns:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupMembershipService.kt
```diff
@@ -101,9 +101,8 @@ class ProgrammeGroupMembershipService(
       attendeeEntity
     }
 
-    // Create appointments in NDelius for each session object for future session only
-    val now = LocalDateTime.now(clock)
-    scheduleService.createNdeliusAppointmentsForSessions(newAttendees.filter { it.session.startsAt > now })
+    // Create appointments in NDelius for each session object
+    scheduleService.createNdeliusAppointmentsForSessions(newAttendees)
```
Jack-Dempster: I'm not sure about this as this would mean we would add any PoP added to a group to all the previous appointments for a group which they may never have attended or supposed to attend

Jack-Dempster: I think probably just skip the scheduling of appointments if there are none in the future

Jack-Dempster (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupMembershipService.kt
pfcairns: Yep, fair point. Will address.

Jack-Dempster (approved):
👍

PR #745, opened 2026-06-18
Return 409 Conflict (not 400) when referral sentence data is stale nDelius

raby:
## What

Returns **HTTP 409 Conflict** instead of **HTTP 400 Bad Request** when an attempt to allocate a referral to a programme group fails because the referral's stored requirement / licence-condition ID no longer exists in nDelius.

## Why

We hit this in production today: two referrals (CRNs `E690304`, `E692471`) couldn't be added to a group. Each was failing inside `ProgrammeGroupMembershipService.validateReferralSentenceDataExistsInNDelius` with a 404 from nDelius — the requirement / licence-condition the referral was originally raised against has been terminated/superseded upstream.

The service was throwing `BusinessException`, which our `ApiExceptionHandler` maps to **400 Bad Request**. That's misleading:
- the request body is well-formed,
- the caller can't fix it by changing the body or retrying,
- the underlying *data state* is what conflicts with the upstream system.

That's exactly what **409 Conflict** is for. It also unblocks the front-end / UX work to render an actionable message rather than a generic "bad request" toast — branching on `409` cleanly separates this case from real validation 400s.

## What changes

`src/main/kotlin/.../service/ProgrammeGroupMembershipService.kt`
- In `validateReferralSentenceDataExistsInNDelius`, the `ClientResult.Failure.StatusCode` branch now throws `ConflictException` instead of `BusinessException`.
- Added a comment explaining the 409 choice.

`src/test/kotlin/.../service/ProgrammeGroupMembershipServiceTest.kt`
- Renamed and updated the two existing tests for `REQUIREMENT` and `LICENCE_CONDITION` to assert `ConflictException`.

No changes to:
- `ApiExceptionHandler` — `ConflictException` is already mapped to `HttpStatus.CONFLICT` with `userMessage = "Conflict: …"`.
- The user-facing message text — kept identical so any consumer reading `userMessage` / `developerMessage` continues to work.
- Telemetry — same `Referral.allocate-to-group.ndelius-validation-failure` event with the same dimensions.
- Sentry — `ConflictException` is captured by the existing handler.

## Out of scope (deliberately)

- The other validation paths in the same method (`null eventId`, `null sourcedFrom`, `ClientResult.Failure.Other` network failures) still throw `BusinessException` → 400. They're not data-drift cases and are reasonable follow-ups in their own right (network failure could become 503).
- UI changes to render a user-friendly message — being handled separately by design.

## Behavioural change for clients

| Scenario | Before | After |
| --- | --- | --- |
| Stored requirement / LC id 404s in nDelius | `400 Bad request: Cannot allocate referral to group: …` | `409 Conflict: Cannot allocate referral to group: …` |
| Everything else | unchanged | unchanged |

The response body shape (`ErrorResponse`) is identical; only the `status` code and `userMessage` prefix change.

## Testing

- `./gradlew test --tests "*ProgrammeGroupMembershipServiceTest"` → 5/5 pass, including the two updated tests.
- ktlint pre-commit hook clean.


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupMembershipService.kt
thomaswilsonxyz, line 277: we should probably update this to make it a little more distinct from what's below on ln. `304`

thomaswilsonxyz (approved):
one small comment but happy for that to be a follow-up one liner PR.

not sure what this will meaningfully change for our users - does our UI handle these any differently ?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupMembershipService.kt
raby: Good call on the duplicate telemetry name, moving that + the controller-level integration tests into a follow-up so this stays minimal

raby:
Thank you. The status code split is allows the UI to display a different message cleanly if they wish.  And for us perusing logs. I agree at the moment today no different to the user.

PR #746, opened 2026-06-18
feat: Also source a User's region from a database table

thomaswilsonxyz:
> [!IMPORTANT]
> This PR contains AI-generated code.  AI prompt is at the bottom of this PR description.  Prompt was used to generate a plan, the plan was reviewed by hand, no significant changes were requested.  I (Wilson) have reviewed all the code of it by hand, and have run the API locally.  I do not believe we can achieve this feature change in a much more simple way than what has achieved.

**WHAT**

This PR introduces the ability to manually (through DB access) override the Regions that a User has access to.

**WHY**

As part of the roll-out of M&D, we are experiencing logistical difficulty with getting our product and development team access to regions.  To do so, we must request roles and teams through the nDelius team, which can be slow and prevent rapid assessment and debugging.

**HOW**

Previously, the User's region (singular) was determined from the nDelius API integration, finding their first team, and the region that team belongs to.  That was used when finding both Referrals and Programme Groups.

This PR introduces two changes:

1. Changes the User:Region relationship from 1-1 to 1-n, in both Programme Group and Referral fetching contexts; and
2. In addition to sourcing Regions from the nDelius client, a new db table (`user_region_override`) and associated entity, fetches the user's regions.

To introduce this change, we require additional database tables and entities, and the `UserService::getUserRegionNames` to fetch the Regions (possibly 0-n), and relevant changes to components of both the `ReferralCaseListService` and `ProgrammeGroupService` to handle the previous single region name (`regionName: String?`) now becoming plural (`regionNames: List<String>`).

A variety of repository components (`ProgrammeGroup`, `ReferralReportingLocation`) have also been updated, which use for straight-forward SQL queries, as opposed to the Criteria-based query selection we have to use for the Referral Case List view (which uses a SQL Materialized View under the hood)

**TESTING**

Integration- and Controller-level testing have been updated to indicate these changes.

---

## AI Prompt

Problem: 

The "Referral Caselist View" currently only shows people Referrals which are in the same region they are in (i will detail how later).  However, as part of the roll-out of our software, we often need people to be in multiple regions at once, and also to add/remove people within our own remit.  Presently, there is an administrative process to re-allocate people to "teams" in an external system, which takes days/weeks, as it is intended for operational staff, not dev staff.  The problem we are facing is three fold: 1) change the relationship of user-region from 1-1 to 1-n (a person can be in many regions), and 2) change the way Referrals and ProgrammeGroups are retrieved from "for one region" to "for a list of regions", and 3) source a user's region from an external service _and_ from our internal database.

Required changes:

1. A new database model (you are to decide the name) which models a manual addition of a user to a region.  This is to have a `username` (which will be a HMPPS Auth Username) and a `region_name`, which will map to a `referral_delivery_location.region_name`.  
2. Changing of the implementation of `ReferralCaselistService::getReferralCaseListItemServiceByCriteria` to take `regionName` from the current source (`getFirstUserRegionDescription`), as well as region_names from the above-mentioned database rows.  This will likely also involve changing of the `UserService::getUserRegions` method, to query this new database table above
3. Changes to the database criteria which search based on `referral_reporting_location.region_name` to match based on an incoming list.  Do the same for programme groups
4. We'll also need to make changes to the Group List view (see the `ProgrammeGroupService:::getProgrammeGroupsForRegion` method), trace the ProgrammeGroupsController for this endpoint

Domain context:

Our users are managed through a central service, called nDelius.  To retrieve the list of Regions available to a User we have gone to nDelius, retrived a list of Teams for the user (`NDeliusIntegrationApiClient::getTeamsForUser`) and worked backwards from teams -> Region.  However the process of assigning someone to a team in nDelius is slow and administrative, and is not appropriate for the fast-scale changes we need during the rollout of new software, and where we may wish to add someone to multiple regions. 

We therefore need a way that we (the product/dev team) can easily add someone to a reason.  This should _only_ be for internal purposes, and shouldn't be long lived.

We'd want a simple, sustainable way to do this, it'll be database only as we don't want/need a UI, but we would like it documented (so that future developers can make changes).

The idea that someone can be in multiple Regions was initially thought not-possible, but as we are learning, there are possible edge cases, so we would want to possibly allow it.  So changing the 1-1 relationship from user to region is a thing to take seriously.  The manual override is less rigorous.


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
```diff
@@ -679,7 +679,11 @@ class ProgrammeGroupService(
   ): GroupsByRegion {
     val groupCohort = if (cohort.isNullOrEmpty()) null else ProgrammeGroupCohort.fromString(cohort)
 
-    val firstUserRegionDescription = userService.getFirstUserRegionDescription(username)
+    val userRegionNames = userService.getUserRegionNames(username)
+    if (userRegionNames.isEmpty()) {
+      throw NotFoundException("Cannot find any regions (or teams) for user $username")
+    }
```
thomaswilsonxyz, line 685: Do we want to throw an Error here, or simply return an empty list?

pfcairns: The UI is effectively "broken" without a user region, so IMHO flagging this as an exception is fine.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/repository/ProgrammeGroupRepository.kt
```diff
@@ -16,16 +16,18 @@ interface ProgrammeGroupRepository :
...
   @Query(
     "SELECT DISTINCT p.probationDeliveryUnitName FROM ProgrammeGroupEntity p " +
-      "WHERE p.regionName = :regionName " +
-      "AND p.probationDeliveryUnitName IS NOT NULL",
+      "WHERE p.regionName IN (:regionNames) " +
+      "AND p.probationDeliveryUnitName IS NOT NULL " +
+      "ORDER BY p.probationDeliveryUnitName",
   )
-  fun findDistinctProbationDeliveryUnitNames(regionName: String): List<String>
+  fun findDistinctProbationDeliveryUnitNames(regionNames: Collection<String>): List<String>
```
pfcairns, line 23: Could the `Collection` just be a `List` instead? Does using a `Collection` give us anything?

pfcairns (approved):
LGTM. Just a very minor comment.

raby (commented):
Great

PR #751, opened 2026-06-19
Apg 1753/stop 400 error in ndelius on session status update

miroslav25a:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
```diff
@@ -447,6 +447,10 @@ class SessionService(
       NotFoundException("Session not found with id: $sessionId")
     }
 
+    if (LocalDateTime.now().isAfter(session.startsAt.plusDays(1))) {
```
DanEgan1: Could this potentially cause issues when local time goes from GMT to BST or vice versa potentially making the session starts at it a different timezone. Might already be handled elsewhere though

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionServiceTest.kt
```diff
@@ -84,6 +85,7 @@ class SessionServiceTest {
       authenticationUtils,
       userService,
       regionService,
+      Clock.systemDefaultZone(),
     )
```
raby, line 94: A bit verbose but there could exist a unit test covering the new 24-hour rejection path. 

All existing `saveSessionAttendance` tests use `SessionFactory` which defaults `startsAt = LocalDateTime.now()`, so the throw is never exercised at unit level. A test using a fixed clock would confirm the behaviour directly without relying solely on the integration test, e.g.:

  @Test
  fun `should throw BusinessException when recording attendance more than 24 hours after session starts`() {
      val fixedClock = Clock.fixed(Instant.now(), ZoneId.systemDefault())
      service = SessionService(..., clock = fixedClock)

      val sessionEntity = SessionFactory()
          .withStartsAt(LocalDateTime.now(fixedClock).minusHours(25))
          ...produce()

      every { sessionRepository.findById(any()) } returns Optional.of(sessionEntity)

      val exception = assertThrows<BusinessException> {
          service.saveSessionAttendance(sessionId, sessionAttendance)
      }
      assertTrue(exception.message!!.contains("after 24 hours"))
  }

raby (commented):
Looks good

raby (approved):
Looks good

PR #752, opened 2026-06-22
APG-2383: Reject invalid referral status transitions

raby:
ReferralService.updateStatus previously looked up the configured transition between the current and requested status but only used the result to decide whether to remove a group membership. If no transition row existed it still saved the requested status verbatim, which let callers move a referral into a state the UI never offers (e.g. the Scheduled -> Programme complete jump reported on M209500).

Throw a BusinessException (HTTP 400 via ApiExceptionHandler) when the requested transition is not configured, while still allowing same-state no-op writes so prior behaviour is preserved for that edge case.

- Add a regression integration test covering Scheduled -> Programme complete, asserting the status history and group membership are untouched.
- Update the concurrent multi-endpoint test to target a configured transition (Awaiting allocation) instead of Awaiting assessment -> On programme, which was never configured.
- Switch the createReferralAndUpdateStatus test helper to persist the status history directly so fixtures can place a referral in any state without being blocked by the new guard.

Before adding any speculative 406-handling, I ran a focused App Insights investigation across dev, preprod, and prod, time-bounded to App Insights' 90-day retention window:

search "M209500" returns no rows for cloud_RoleName == "hmpps-accredited-programmes-manage-and-deliver-api" or …-ui in any environment. The incident therefore predates 24 March 2026 and is outside retention.

requests | where resultCode == "406" on prod for our API: 0 rows in 90 days.
requests | where resultCode == "406" on prod for our UI BFF: 0 rows in 90 days.
exceptions | summarize by type on prod: 0 HttpMediaTypeNotAcceptableException, HttpMessageNotWritableException, or any ContentNegotiation* class in 90 days.

The framework-content-negotiation hypothesis is not supported by 90 days of production data. This PR therefore ships only the unconfigured-transition guard (the half of the bug we can prove and test) and does not add a speculative 406 handler.


src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/utils/TestReferralHelper.kt
```diff
@@ -282,7 +287,17 @@ class TestReferralHelper {
...
+    // state without being constrained by configured transitions (the production
+    // updateStatus path validates transitions).
+    referralStatusHistoryRepository.save(
+      ReferralStatusHistoryEntity(
+        referral = referral,
+        referralStatusDescription = status,
+        additionalDetails = null,
+        createdBy = "AUTH_USER",
+      ),
+    )
```
Jack-Dempster, line 300: only thing with removing this is this means referrals using this helper can be put into states that would not be possible through using the service normally. I've tried to avoid doing this in this class so we would have realistic referral data but if it is providing a big blocker to tests then this makes sense

raby: Fair callout, Jack. I went back and forth on this one. The reason for the direct save: several calls use this helper to land a referral straight in a late-stage status (PROGRAMME_COMPLETE, AWAITING_ALLOCATION from initial, etc.) that isn't reachable in one configured transition. 

Pre-APG-2383 the service path silently allowed it; the new transition guard (correctly) doesn't.

Options I considered:

1: Have the helper walk the transitions itself, work out the shortest valid path from the initial status to whatever the test wants, and step through it. Doable, but it ties every fixture to the contents of the referral_status_transition table. Tweak a row in a migration and you'd get unrelated tests failing for reasons that have nothing to do with what they're testing. Felt like a trap.

2. Rewrite the affected tests to walk the transitions themselves through the service, a lot of changes across files, and most of those tests don't care about transitions at all. They just need a referral that happens to be in PROGRAMME_COMPLETE so they can assert something else.

3. What I did, write straight to the status history for fixture setup, and leave updateReferralStatus (which still goes through the service) for the tests that actually exercise transition behaviour. The helper's job is "give me a referral in state X", instead of "simulate how it got there".

Went with (3) on the basis that the helper's job is fixture setup. The inline comment is there to flag the asymmetry to the next reader.

That said, your point about the name still suggesting it goes through the service is a good one. I've raised a follow-up to rename it to something like `createReferralWithStatusHistory` and tighten the KDoc so the bypass is obvious at the call site: APG-2283. Keeping it out of this PR to avoid mixing test-helper hygiene with the production code change.

Jack-Dempster: cool. I'm happy if there are lots of tests that would rely on this behaviour then we can leave as is. Thanks for the explanation

Jack-Dempster (approved):

PR #754, opened 2026-06-24
APG-2390 Fix Rescheduling algorithm

Jack-Dempster:
## WHAT

When rescheduling sessions in a group

via Slots change
- Ensure only future sessions are affected by the slot change

via Session Date and time change
- Ensure that future sessions being rescheduled match the module and session order rather than using the session date as the reschedule point use now instead


niaevans9 (approved):
lgtm

pfcairns (approved):
👍

raby (commented):
LGTM

PR #756, opened 2026-06-25
fix: Find the Users for the Waitlist by the *group's* region, not the currently logged-in user's region

thomaswilsonxyz:
**WHAT**

This commit changes the way that the Waitlist on a Programme Group finds Referrals - we now use the Group's region rather than one from the currently-logged in user.


niaevans9 (approved):
lgtm

raby (commented):
Looks good

PR #758, opened 2026-06-25
Apg 1865/enrich referral personal details

miroslav25a:


raby (dismissed):

src/main/resources/application.yml
```diff
@@ -37,7 +37,8 @@ spring:
 
   datasource:
     hikari:
-      connectionTimeout: 1000
+      maximumPoolSize: 30
```
pfcairns: This looks like it should be: `spring.datasource.hikari.maximum-pool-size`

From: https://docs.spring.io/spring-boot/appendix/application-properties/index.html#application-properties.data.spring.datasource.hikari

src/main/resources/application.yml
```diff
@@ -37,7 +37,8 @@ spring:
...
+      connectionTimeout: 10000
```
pfcairns: And this : `spring.datasource.hikari.connection-timeout`

pfcairns (changes requested):
The config looks like it needs tweaking slightly.

src/main/resources/application.yml
miroslav25a: comment addressed

src/main/resources/application.yml
miroslav25a: comment addressed

pfcairns (approved):
Thanks for making the changes

raby (commented):
Looks good

PR #760, opened 2026-06-26
Apg 1865/prevent null pointer exception when latestReferralStatus is null

miroslav25a:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/model/ReferralDetails.kt
```diff
@@ -177,26 +181,26 @@ data class ReferralDetails(
...
       cohort = cohort ?: OffenceCohort.GENERAL_OFFENCE,
       hasLdc = LdcStatus.fromBoolean(hasLdc).value,
       hasLdcDisplayText = LdcStatus.getDisplayText(hasLdc),
-      currentStatusDescription = latestReferralStatus.description,
+      currentStatusDescription = latestReferralStatus?.description ?: DEFAULT_STATUS_DESCRIPTION_VALUE,
       currentlyAllocatedGroupCode = currentlyAllocatedGroup?.programmeGroup?.code,
       currentlyAllocatedGroupId = currentlyAllocatedGroup?.programmeGroup?.id,
-      pdu = referral.referralReportingLocation?.pduName ?: "Unknown",
-      reportingTeam = referral.referralReportingLocation?.reportingTeam ?: "Unknown",
+      pdu = referral.referralReportingLocation?.pduName ?: DEFAULT_VALUE,
```
raby, line 202: We could have DEFAULT_PDU_VALUE, DEFAULT_TEAM_VALUE? To help future

pfcairns (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/model/ReferralDetails.kt
miroslav25a: there are no specific default values for PDU and TEAM thats why I moved the generic value to the generic constant

PR #771, opened 2026-07-01
full stop added and hypen removed from design review

Alison-Weatherston:
General offence LDC and Sexual offence LDC - hyphens removed


raby (approved):
Looks good!

PR #774, opened 2026-07-02
APG-2461 Programme completion event not being sent from finding incorrect session

Jack-Dempster:
If there were multiple schedule/attended post programme review sessions there was a chance that the wrong session could be being looked at for attendance and therefore the termination event would never publish.

This is because we previously had
```kotlin
val postProgrammeReviewSession = currentGroupMembership.programmeGroup.sessions
  .find { it.moduleSessionTemplate.module.isPostProgrammeModule() && !it.isPlaceholder } 

```
which could non-deterministically find the post programme review session for a group.


raby (approved):
Looks great

PR #775, opened 2026-07-03
APG-2377: detect nDelius terminated requirement 400s

raby:
# APG-2377 — Detect "linked requirement terminated in nDelius" (part 1 of N)

## What this PR does

Adds detection + telemetry for the specific nDelius failure mode where the linked
requirement (or licence condition) has been terminated in nDelius. **No user-visible
behaviour change.** Nothing is auto-closed, nothing is retried, nothing new appears
on the UI.

The failure signature — verified across dev / preprod / prod App Insights — is:

```
POST /appointments  →  HTTP 400
Body:  {"status":400,"message":"Invalid Requirement IDs: [<id>]"}
```

When this specific shape is detected inside
`ScheduleService.createNdeliusAppointmentsForSessions`, we now:

1. Fire an additional App Insights event
   `Appointment.create-nDelius.terminated-requirement` with dimensions
   `{integrationActionType, outcome, requirementIds, crns, eventNumbers, groupId}` —
   in addition to the existing `Appointment.create-nDelius.failure` event, which
   still fires for **every** failure. Watching both lets us slice terminated-requirement
   failures out of the general failure count without regex-matching messages.
2. Throw a new `TerminatedRequirementException(requirementIds, message, cause)` that
   `extends BusinessException`, carrying the parsed requirement IDs. Because it IS-A
   `BusinessException`, the existing global exception handler routes it to the same
   HTTP 500 mapping as before — **no behaviour change for the user in this PR**.

## Why now, why so small

The team is having ongoing discussions about the bigger question ("what should the
user actually see?" — better error page? auto-close? product redesign?). This PR
deliberately doesn't pre-empt any of those. It only lands the **observation seam**
so:

- We can measure the true incidence rate in App Insights immediately.
- Whichever direction the team goes, the branching point (typed exception) is already
  in place — future work catches `TerminatedRequirementException` at whatever
  boundary it decides on, without touching `ScheduleService` again.

## Scope — deliberately included

- New `TerminatedRequirementDetector` — pure object with `isTerminated(body)` and
  `extractRequirementIds(body)`. Fully unit-tested in isolation.
- New `TerminatedRequirementException : BusinessException` carrying the parsed IDs.
- One-word change: `BusinessException` is now `open` (was implicitly `final`) so
  it can be subclassed. This matches its conceptual role as a family of business
  exceptions — no other code needed to change.
- Detection branch inside `ScheduleService.createNdeliusAppointmentsForSessions`,
  placed after the existing `.failure` telemetry and before the existing generic
  `throw` (so general failure counts remain intact).
- Unit tests (`TerminatedRequirementDetectorTest`, `ScheduleServiceTest`) covering:
  matching 400 → typed exception + both telemetry events; non-matching 400 → plain
  `BusinessException` only; non-400 status → plain `BusinessException` only.
- Integration test in `ScheduleServiceIntegrationTest` — real WireMock returning
  the terminated 400 body → typed exception surfaces with the correct
  `requirementIds`. New helper `nDeliusApiStubs.stubTerminatedRequirementPostAppointmentsResponse()`.

## Scope — deliberately OUT (for now)

Each of these is a valid follow-up and none of them are blocked by this PR:

- ❌ **No status changes / no auto-close.** The user still sees the same error the
  UI shows today when appointment creation fails. Whether to auto-close, prompt,
  or redesign the flow is a live product discussion.
- ❌ **No new response fields / no controller changes.** `AllocateToGroupResponse`
  is untouched. No API contract shift.
- ❌ **No feature flag.** Nothing gates user-visible behaviour because there IS no
  user-visible behaviour change — just telemetry + a more precise exception type.
  Adding a flag now would be ceremony without a payoff.
- ❌ **No schema migration.** Nothing persisted.
- ❌ **No F&R domain events.** Nothing new emitted downstream.
- ❌ **No auto-close-service / coordinator layer / `AuditorContext` handling.**
  That's the "part 2" design (see `docs/APG-2377-terminated-requirement-design.md`
  v5) if the team decides to go that direction. Not needed for this PR.
- ❌ **No mid-programme detection.** If a requirement is terminated after allocation,
  this PR doesn't catch it. Would need a nDelius domain-event subscription or a
  reconciliation job — separate ticket.
- ❌ **No 404 auto-close.** The `"Requirement with id of X not found"` failure is a
  different class (wrong stored eventId / deleted requirement) — separate follow-up.
- ❌ **No handler added for `BusinessException`** in
  `AccreditedProgrammesManageAndDeliverApiExceptionHandler`. Doing so would improve
  the pre-flag UX for every `BusinessException`, not just this one — scope-widening.
- ❌ **No alerting** on `terminated-requirement` telemetry drop-to-zero. Follow-up
  once we have baseline data.

## Risk register

1. **False-negative on wording change** — if nDelius changes the "Invalid Requirement
   IDs" phrasing, detection silently stops. Mitigation: the generic
   `Appointment.create-nDelius.failure` event still fires for every failure, so
   watching the ratio (terminated / total-failure) drops to zero against a stable
   or rising baseline is the tripwire. A `KQL` alert is a cheap follow-up. This
   trade-off is documented as a code comment on the `MARKER` constant in
   `TerminatedRequirementDetector` — deliberately kept short (one paragraph) rather
   than sprinkled throughout the code.
2. **Downstream `catch (BusinessException)` behaviour** — because
   `TerminatedRequirementException extends BusinessException`, any existing
   `catch (BusinessException)` clauses upstream continue to work unchanged.
   Verified: no such catches exist in the call chain from
   `ProgrammeGroupController.allocateReferralToGroup` (the exception falls straight
   through to the global handler exactly as before). Making `BusinessException`
   `open` is safe.

## Testing done

| Layer | File | Notes |
|---|---|---|
| Unit — pure | `TerminatedRequirementDetectorTest` (14 tests) | Marker match/no-match, ID extraction (single / multi / whitespace-padded / empty / missing brackets / null / no marker) |
| Unit — service | `ScheduleServiceTest` (3 tests) | Terminated body → typed exception + both telemetry events; non-matching 400 → plain `BusinessException` + `.failure` telemetry only; 503 → plain `BusinessException`, no `.terminated-requirement` telemetry even if body would match |
| Integration | `ScheduleServiceIntegrationTest` (1 new test) | End-to-end via `programmeGroupMembershipService.allocateReferralToGroup` with WireMock stubbed to return the terminated 400 body — verifies the typed exception surfaces at the boundary with the correct `requirementIds` |

All 20 new/edited tests pass locally.

## Files changed

**Prod (3):**
- `service/TerminatedRequirementDetector.kt` — NEW
- `common/exception/TerminatedRequirementException.kt` — NEW
- `common/exception/BusinessException.kt` — `class` → `open class` (one word)
- `service/ScheduleService.kt` — detection branch inside existing
  `Failure.StatusCode` branch of `createNdeliusAppointmentsForSessions`

**Test (3):**
- `service/TerminatedRequirementDetectorTest.kt` — NEW
- `service/ScheduleServiceTest.kt` — NEW
- `integration/wiremock/stubs/NDeliusApiStubs.kt` — new
  `stubTerminatedRequirementPostAppointmentsResponse(requirementId)` helper
- `service/ScheduleServiceIntegrationTest.kt` — new integration test

## App Insights query to eyeball after merge

```kusto
customEvents
| where name in ("Appointment.create-nDelius.failure",
                 "Appointment.create-nDelius.terminated-requirement")
| summarize count() by name, bin(timestamp, 1d)
| render timechart
```

Once deployed, the `terminated-requirement` line should track a subset of the
`failure` line. If the ratio suddenly drops to zero while `failure` stays stable
or rises, that's the signal that the nDelius wording has changed.


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/common/exception/TerminatedRequirementException.kt
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Thrown when nDelius rejects appointment creation because the linked requirement
+ * (or licence condition) has been terminated in nDelius.
+ *
+ * Extends [BusinessException] so today it flows through the existing global
+ * exception handler unchanged (generic `Exception → HTTP 500` — same as any other
+ * BusinessException). Callers that want to react to this specific failure can catch
+ * the typed exception; everything else keeps working exactly as before.
+ */
+class TerminatedRequirementException(
```
pfcairns: Questioning whether we need this additional exception, or whether reusing `BusinessException` with a pertinent message might suffice.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ScheduleService.kt
```diff
@@ -461,6 +463,33 @@ class ScheduleService(
             "responseBody" to (response.body?.take(500) ?: ""),
           ),
         )
+
+        // APG-2377: detect the specific "linked requirement terminated in nDelius" case.
```
pfcairns: We shouldn't really store JIRA ticket numbers in our code. The JIRA ticket will be in the branch name, and github history.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ScheduleService.kt
```diff
@@ -461,6 +463,33 @@ class ScheduleService(
...
+        // Fires an additional, finer-grained telemetry event and throws a typed exception
+        // so future code can react to this failure specifically. Behaviour today is
+        // unchanged for the caller — TerminatedRequirementException IS-A BusinessException.
+        if (response.status.value() == HttpStatus.BAD_REQUEST.value() &&
+          TerminatedRequirementDetector.isTerminated(response.body)
```
pfcairns: could just be : `response.body.contains("Invalid Requirement IDs")` which negates the need for the additional  `TerminatedRequirementDetector`  class, and is obvious in it's behaviour and intention.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/TerminatedRequirementDetector.kt
```diff
@@ -0,0 +1,30 @@
...
+ * drop-to-zero in the terminated-requirement dimension against a stable failure count
+ * is the tripwire.
+ */
+object TerminatedRequirementDetector {
+  private const val MARKER = "Invalid Requirement IDs"
+  private val ID_LIST_REGEX = Regex("""Invalid Requirement IDs:\s*\[([\d,\s]*)]""")
+
+  fun isTerminated(body: String?): Boolean = body?.contains(MARKER) == true
+
+  fun extractRequirementIds(body: String?): List<String> = ID_LIST_REGEX.find(body ?: "")
```
pfcairns: A nice to have, but probably unnecessary.

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ScheduleServiceIntegrationTest.kt
```diff
@@ -632,4 +634,33 @@ class ScheduleServiceIntegrationTest(@Autowired private val sessionRepository: S
     // Should be the correct date
     assertThat(nextSlotDate).isEqualTo(LocalDate.of(2126, 7, 10))
   }
+
+  // ------------------------------------------------------------------
+  // APG-2377 — nDelius "linked requirement terminated" detection
```
pfcairns: JIRA ticket number in code which can be removed. 
See "Journal Comments" in this article : https://bytedev.medium.com/code-comment-anti-patterns-and-why-the-comment-you-just-wrote-is-probably-not-needed-919a92cf6758

pfcairns (commented):
Good work - could be simplified in a few places, see comments.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/common/exception/TerminatedRequirementException.kt
raby: Both are reasonable options, I did think this too but thought creating custom exceptions maybe a good way to go for sentry etc.. happy for it to be a BusinessException - simplist solution without too much change

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ScheduleService.kt
raby: Yes

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ScheduleService.kt
raby: Yes simplifies things

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ScheduleServiceIntegrationTest.kt
raby: Good share!

pfcairns (approved):
Thanks for addressing the comments.

PR #776, opened 2026-07-03
Apg 2384/unable to find pop changed crn

miroslav25a:


src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/event/DomainEventsListenerIntegrationTest.kt
```diff
@@ -659,4 +731,176 @@ class DomainEventsListenerIntegrationTest : IntegrationTestBase() {
...
+    testDataGenerator.createGroup(group)
+    val groupMembership =
+      ProgrammeGroupMembershipFactory().withReferral(referralEntity).withProgrammeGroup(group).produce()
+    testDataGenerator.createGroupMembership(groupMembership)
+    val savedReferral = referralRepository.findByCrn(referralEntity.crn)[0]
+    val nDeliusPersonalDetails = NDeliusPersonalDetailsFactory().produce()
+
+    nDeliusApiStubs.stubAccessCheck(granted = true, savedReferral.crn)
+    nDeliusApiStubs.stubPersonalDetailsResponse(nDeliusPersonalDetails)
+    nDeliusApiStubs.stubSuccessfulSentenceInformationResponse(savedReferral.crn, savedReferral.eventNumber)
```
Jack-Dempster: For these tests you could replace this whole block with if we just need a basic referral
```kotlin
val savedReferral = testReferralHelper.createReferral()
```

Jack-Dempster (dismissed):
Looks good! One minor comment for the future but it doesn't stop this being merged.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/event/ReferralMergedHandler.kt
```diff
@@ -0,0 +1,87 @@
...
+        "Referral.merged-event-processed.success",
+        mapOf(
+          APP_INSIGHTS_TARGET_EVENT_TYPE_PROPERTY_KEY to message.eventType,
+          APP_INSIGHTS_SOURCE_CRN_PROPERTY_KEY to sourceCrn,
+          APP_INSIGHTS_TARGET_CRN_PROPERTY_KEY to targetCrn,
+        ),
+      )
+    } catch (e: Exception) {
+      log.error("Error handling ReferralMergedEvent: ${e.message}", e)
+      telemetryClient.logToAppInsights(
```
raby, line 78: The failure telemetry doesn't include the CRNs that were being processed, which makes it hard to debug from App Insights alone. The `sourceCrn`/`targetCrn` variables are in scope here (they've already been validated as non-null before we reach `updateReferralCrn`). Should we add them to the failure map, same as in the success event above?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralService.kt
```diff
@@ -279,6 +279,21 @@ class ReferralService(
...
   }
 
+  /**
+   * Updates the CRN of all referrals associated with the given current CRN to the new CRN.
+   *
+   * @param currentCrn The current CRN associated with the referrals that need to be updated.
+   * @param newCrn The new CRN to replace the current CRN in the referrals.
+   */
+  fun updateReferralCrn(currentCrn: String, newCrn: String) {
+    val referrals = referralRepository.findByCrn(currentCrn)
```
raby, line 289: If `findByCrn` returns empty (we don't hold any referrals for this CRN), this silently succeeds, no log, no telemetry. The caller will fire `.success` telemetry even though nothing changed. This is probably the right behaviour (we may not track every person), but a log line would help with debugging in prod maybe?

```
if (referrals.isEmpty()) {
    log.info("updateReferralCrn: no referrals found for CRN $currentCrn — nothing to update")
    return
}
```

raby (commented):
Looks good overall, just a couple of minor logging improvement ideas

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/event/ReferralMergedHandler.kt
miroslav25a: I don't think we can, the sourceCrn/targetCrn vars are part of the try section which is unaccessible in a catch. Also, I don't think that taking out initialisation of the sourceCrn/targetCrn vars outside the try section is a good idea, as this can produce exceptions. But, what we can do is add the whole SQS message to the telemetry event

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/event/DomainEventsListenerIntegrationTest.kt
miroslav25a: comment addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralService.kt
miroslav25a: comment addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/event/ReferralMergedHandler.kt
raby: Thank you for explaining with a deeper look

raby (approved):
Looks good

PR #777, opened 2026-07-03
APG-2463 : Don't submit session outcome or notes if it has not changed

Jack-Dempster:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
```diff
@@ -468,12 +473,53 @@ class SessionService(
...
     }
 
-    val attendees = sessionAttendance.attendees
-    val sessionAttendanceEntities = getSessionAttendanceFromAttendees(attendees, session)
+    val latestAttendanceByReferralId = session.attendances
+      .groupBy { it.groupMembership.referral.id }
+      .mapValues { (_, attendances) ->
+        attendances.maxWithOrNull(
+          compareBy<SessionAttendanceEntity> { it.createdAt }
+            .thenBy { it.id },
```
raby: The `latestAttendanceByReferralId` uses `thenBy { it.id }` (UUID) as a tiebreaker when two attendance records share the same `createdAt`. UUID comparison is lexicographic, not chronological, so this tiebreaker is arbitrary for same-millisecond writes. In practice this is extremely unlikely to matter, but it's worth noting. Could use `thenBy { it.recordedAt }` as an arguably meaningful secondary sort if `createdAt` granularity is a concern, or just document the tiebreaker intent. (really unlikely!)

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/SessionControllerIntegrationTest.kt
```diff
@@ -2024,6 +2024,123 @@ class SessionControllerIntegrationTest : IntegrationTestBase() {
...
+    assertThat(exception.userMessage)
+      .contains("Attendance outcome cannot be changed once it has been recorded")
+      .contains(attendee.referral.crn)
+    val attendances = sessionRepository.findById(sessionId).get().attendances
+    assertThat(attendances).hasSize(1)
+    assertThat(attendances.first().outcomeType.code).isEqualTo(ATTC)
+  }
+
+  @Test
+  fun `should not update nDelius when resubmitting an unchanged attendance`() {
```
raby, line 2078: There is no unit test covering the case "first submission with no notes, then resubmit with same outcome and no notes" we could add one. The existing `should not save or update ndelius when outcome and notes are unchanged` only covers the "with notes" path. Example:

```kotlin
@Test
fun `should not save or update ndelius when outcome is unchanged and no notes were ever recorded`() {
    val sessionId = UUID.randomUUID()
    val referralId = UUID.randomUUID()
    val referralEntity = ReferralEntityFactory().withId(referralId).withPersonName("John Smith").produce()
    val sessionEntity = sessionWithAttendees(listOf(referralEntity))
    recordAttendanceOnSession(sessionEntity, referralEntity, ATTC) // no notes

    val sessionAttendance = SessionAttendanceFactory()
        .withAttendees(listOf(SessionAttendeeFactory().withReferralId(referralId).withOutcomeCode(ATTC).produce()))
        .produce()

    every { sessionRepository.findById(any()) } returns Optional.of(sessionEntity)

    val result = service.saveSessionAttendance(sessionId, sessionAttendance)

    assertThat(result.responseMessage).isEqualTo("Attendance saved for session $sessionId")
    assertThat(sessionEntity.attendances).hasSize(1)
    verify(exactly = 0) { sessionRepository.save(any()) }
    verify(exactly = 0) { nDeliusIntegrationApiClient.updateAppointmentsInDelius(any()) }
}

raby (dismissed):
Good work

pfcairns (dismissed):

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/SessionControllerIntegrationTest.kt
Jack-Dempster: have added this test now

raby (approved):

PR #778, opened 2026-07-06
APG-2451 Replace group wait list materialized view with standard view 

pfcairns:
### Context

- We've seen various issues of late caused by our use of materialzed views and triggers, such as the one detailed in this JIRA ticket:

  Materialised view "group_waitlist_item_view" fails to refresh with "duplicate rows without any null columns", causing DataIntegrityViolationException on POST /group/{groupId}/allocate/{referralId} and blocking user allocations in prod.

This PR seeks to address this and other issues by converting the group wait list view from a materialized view to a regular view to mitigate this issue. Testing will occur in dev and pre-prod to ensure there is no performance degradation.

_NB. The Custody AcP API uses regular views for displaying almost identical dashboards, and performance has not been a concern._

### How

- Replaced the materialized view `group_waitlist_item_view` with a standard view.
- Removed associated triggers and refresh logic tied to the materialized view.


Jack-Dempster (approved):
This looks good. Good shout on replacing it tbh with the issues we've had with materialised views I think this makes sense

raby (commented):
Wanted to spend some time having a good look, looks good - I agree great approach for our use case, simple

PR #779, opened 2026-07-07
Apg 2443/bus stopped pop added to past sessions

miroslav25a:


DanEgan1 (dismissed):

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ProgrammeGroupControllerIntegrationTest.kt
```diff
@@ -1356,10 +1356,13 @@ class ProgrammeGroupControllerIntegrationTest : IntegrationTestBase() {
...
 
-      // Verify the referral is still added as attendee to all group sessions (past and future)
+      // Verify the referral is added as attendee only to group sessions in the future
       val allGroupSessions = foundReferral.programmeGroupMemberships.first().programmeGroup.sessions
         .filter { it.sessionType == SessionType.GROUP }
-      assertThat(allGroupSessions.sumOf { it.attendees.count() }).isEqualTo(21)
+      assertThat(allGroupSessions.size).isEqualTo(21)
+      val allFutureGroupSessions = foundReferral.programmeGroupMemberships.first().programmeGroup.sessions
+        .filter { it.sessionType == SessionType.GROUP && it.startsAt > now }
+      assertThat(allGroupSessions.sumOf { it.attendees.count() }).isEqualTo(allFutureGroupSessions.size)
```
Jack-Dempster: I think we should be explicit with the assertion here rather than calculating it in the test.

This assertion could pass if one session has duplicated attendees and another session had no attendees but the sum could still be the same.

Maybe adding an assertion that all past sessions have no attendees and all future ones only have 1?

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ProgrammeGroupControllerIntegrationTest.kt
Jack-Dempster, line 1: could we also add a test to verify that we do not call ndelius if there are only past sessions when the PoP is added

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ProgrammeGroupControllerIntegrationTest.kt
miroslav25a: comment addressed

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ProgrammeGroupControllerIntegrationTest.kt
miroslav25a: comment addressed

Jack-Dempster (approved):
looks good mate, thanks for making those changes

PR #782, opened 2026-07-07
APG-2451 Replace remaining materialized views  with standard views 

pfcairns:
Removed materialized views `referral_caselist_item_view` and `reporting_group_size`, replacing them with standard views. 
Also removed associated triggers, refresh logic, and dependencies.


miroslav25a (approved):

raby (approved):
Looks good - Both views use `ROW_NUMBER() OVER (PARTITION BY ...) ` functions that were previously precomputed as materialised views. The caselist endpoint in particular looks could be called frequently, worth confirming query performance is acceptable in prod as a test during the team test column phase?

PR #783, opened 2026-07-08
APG-2451 Refactor views to use joins

pfcairns:
Refactor existing views by replacing subqueries with SQL joins to improve query performance and maintainability.
Added indexes for referral caselist item view.
No functional changes have been introduced.


raby (approved):
Looks good!

PR #786, opened 2026-07-09
Apg 2399/refresh referral details from ndelius

miroslav25a:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJob.kt
```diff
@@ -0,0 +1,119 @@
...
+      val referralPage: Page<ReferralEntity> = referralService.findAllByUpdatedAtBefore(cutoffTime, pageable)
+      val referrals: MutableList<ReferralEntity> = referralPage.getContent()
+
+      referrals.forEach { referral ->
+        val referralId = referral.id!!
+        val message = getDomainEventMessage(referralId, referral.crn)
+        log.info("Publishing ${ACP_M_AND_D_REFERRAL_DETAILS_UPDATED.value} event for referralId: $referralId")
+        publishEvent(message, referralId)
+        totalUpdated++
+      }
```
raby, line 62: `publishEvent` catches, logs to AppInsights, then re-throws. A single SNS publish failure will throw out of the `forEach`, exit the `while` loop, and abort `process()`. 

The existing `ReferralEventService.publishEvent `pattern also re-throws, but that's called once per HTTP request (not in a batch loop), so it's fine there. This is genuinely different because it's inside a paginated loop.

example:
```
`// ReferralDetailsUpdatedJob.kt
referrals.forEach { referral ->
    val referralId = referral.id!!
    val message = getDomainEventMessage(referralId, referral.crn)
    log.info("Publishing ${ACP_M_AND_D_REFERRAL_DETAILS_UPDATED.value} event for referralId: $referralId")
    try {
        publishEvent(message, referralId)
        totalUpdated++
    } catch (exception: Exception) {
        log.error("Failed to publish event for referralId: $referralId, skipping and continuing", exception)
    }
}`
```

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJob.kt
```diff
@@ -0,0 +1,119 @@
...
+    var totalUpdated = 0
+    val cutoffTime = LocalDateTime.now(clock).minusSeconds(cutoff)
+    val startingPage = 1
+    var currentPage = startingPage
+    var hasMoreRecords = true
+
+    while (hasMoreRecords) {
+      val pageable: Pageable = PageRequest.of(currentPage - 1, pageSize)
+      val referralPage: Page<ReferralEntity> = referralService.findAllByUpdatedAtBefore(cutoffTime, pageable)
+      val referrals: MutableList<ReferralEntity> = referralPage.getContent()
```
raby: A small one. 

`.getContent() `is Java-style getter call in code.

Can use kotlin?
`val referrals = referralPage.content`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/event/ReferralDetailsUpdatedHandler.kt
```diff
@@ -0,0 +1,88 @@
...
+        mapOf(
+          APP_INSIGHTS_TARGET_EVENT_TYPE_PROPERTY_KEY to message.eventType,
+          APP_INSIGHTS_REFERRAL_ID_PROPERTY_KEY to referralIdString,
+        ),
+      )
+
+      messageHistoryRepository.save(message.toEntity(objectMapper.writeValueAsString(message)))
+      val referralId = UUID.fromString(referralIdString)
+      runBlocking {
+        referralService.refreshPersonalDetailsForReferral(referralId, false)
```
raby, line 65: when `refreshPersonalDetailsForReferral` returns `null` (referral UUID doesn't exist in DB), the handler currently logs "Ending handle..." and fires "Referral.details-updated-event-processed.success" potentially making a missing referral indistinguishable from a real success in AppInsights. Could use:

```
`runBlocking {
    val result = referralService.refreshPersonalDetailsForReferral(referralId, false)
    if (result == null) {
        log.warn("No referral found for referralId: $referralId — skipping refresh")
        telemetryClient.logToAppInsights(
            APP_INSIGHTS_PROCESSED_FAILURE_EVENT_NAME_PROPERTY_VALUE,
            mapOf(
                APP_INSIGHTS_ERROR_MESSAGE_PROPERTY_KEY to "referral not found",
                APP_INSIGHTS_REFERRAL_ID_PROPERTY_KEY to referralId.toString(),
            ),
        )
        return@runBlocking
    }
}`
```

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/event/ReferralDetailsUpdatedHandler.kt
```diff
@@ -0,0 +1,88 @@
...
+      "Referral.details-updated-event-processed.failure"
+  }
+
+  fun handle(sqsMessage: SQSMessage) {
+    val messageId = sqsMessage.messageId
+    log.info("Starting handle for messageId: $messageId")
+    try {
+      val message: DomainEventsMessage = objectMapper.readValue<DomainEventsMessage>(sqsMessage.message)
+      val referralIdString = message.referralId
+      if (referralIdString.isNullOrEmpty()) {
```
raby, line 42: There is no test I think covering this branch? Could add test:

LLM generated:
```
`@Test
fun `should not process referral details updated event when referralId is blank`() {
    // Given
    val eventType = "accredited-programmes-manage-and-deliver.referral.details-updated"
    val domainEventsMessage = DomainEventsMessageFactory()
        .withAdditionalInformation(mapOf("referralId" to ""))  // blank referralId
        .withEventType(eventType)
        .produce()

    // When
    domainEventsQueueConfig.sendDomainEvent(domainEventsMessage)

    // Then
    await withPollDelay ofMillis(100) untilCallTo {
        with(domainEventsQueueConfig) { domainEventQueue.countAllMessagesOnQueue() }
    } matches { it == 0 }

    // No message history saved, no referral updated
    assertThat(messageHistoryRepository.findAll()).isEmpty()
}`
```

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJobIntegrationTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    } matches { it != null }
+
+    messageHistoryRepository.findAll().first().let {
+      assertThat(it.id).isNotNull
+      assertThat(it.eventType).isEqualTo("accredited-programmes-manage-and-deliver.referral.details-updated")
+      assertThat(it.detailUrl).isEqualTo("$madBaseUrl/referral-details/$referralId/personal-details")
+      assertThat(it.description).isEqualTo("An Accredited Programmes referral details in community have been updated.")
+      assertThat(it.message).contains("\"additionalInformation\":{\"referralId\":\"$referralId\"}")
+    }
+  }
```
raby, line 63: we could add a test for a case where no referrals are stale so the job does nothing?


LLM generated:
```
`@Test
fun `should not publish events when no referrals are stale`() {
    // Given — referral created just now, updatedAt is fresh (within the 1 second cutoff)
    testReferralHelper.createReferral()

    // When — run immediately, no delay, so updatedAt is NOT older than cutoff
    job.process()

    // Then — no events published, no message history created
    assertThat(
        with(domainEventsQueueConfig) { domainEventQueue.countAllMessagesOnQueue() }
    ).isEqualTo(0)
    assertThat(messageHistoryRepository.findAll()).isEmpty()
}`
```

src/main/resources/application-local.yml
```diff
@@ -46,7 +46,7 @@ hmpps:
         queueName: hmpps_domain_events_queue
         dlqName: hmpps_domain_events_dlq
         subscribeTopicId: hmppseventtopic
-        subscribeFilter: '{"eventType":[ "interventions.community-referral.created", "interventions.community-referral.imported", "probation-case.merge.completed", "probation-case.unmerge.completed"] }'
+        subscribeFilter: '{"eventType":[ "interventions.community-referral.created", "interventions.community-referral.imported", "probation-case.merge.completed", "probation-case.unmerge.completed", "accredited-programmes-manage-and-deliver.referral.details-updated"] }'
```
pfcairns, line 49: does the additional filter need to be in the base `application.yml` or is it just for local testing?

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJobIntegrationTest.kt
```diff
@@ -0,0 +1,64 @@
...
+
+  @Test
+  fun `should process referral details updated job`() {
+    // Given
+    val savedReferral = testReferralHelper.createReferral()
+    val referralId = savedReferral.id
+
+    // When
+    runBlocking {
+      delay(2000) // this is required for savedReferral.updatedAt to be older than 1 second from now
```
pfcairns: Can we reduced this to some millisecond value as it will add 2 seconds to every run of the full test suite, local and on the Ci/cd pipeline.

src/test/resources/application-test.yml
```diff
@@ -8,6 +8,13 @@ management.endpoint:
 hmpps-auth:
   url: "http://localhost:9999/auth"
 
+app:
+  scheduling:
+    referral-details-updated:
+      cron: "0 0 1 * * *"  # every day at 1 AM
+      chunk-size: 100  # process referral records in chunks
+      cutoff: 1 # before updatedAt in seconds
```
pfcairns, line 16: `cutoff: 0.1` ? 😄

pfcairns (commented):
Looks good, just a couple of minor comments.

src/main/resources/application-local.yml
miroslav25a: just for our local testing. it needs to be added in the terraform, cloud platform envs the sqs-domain-events.tf files

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJobIntegrationTest.kt
miroslav25a: it can go down to 1001, I don't think it can go below a second. will make a change

src/test/resources/application-test.yml
miroslav25a: its an Int value in the class, don't think it can handle below 1

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJob.kt
miroslav25a: comment addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/event/ReferralDetailsUpdatedHandler.kt
miroslav25a: comment addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/event/ReferralDetailsUpdatedHandler.kt
miroslav25a: comment addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJob.kt
miroslav25a: comment addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJob.kt
```diff
@@ -0,0 +1,124 @@
...
+    while (hasMoreRecords) {
+      val pageable: Pageable = PageRequest.of(currentPage - 1, pageSize)
+      val referralPage: Page<ReferralEntity> = referralService.findAllByUpdatedAtBefore(cutoffTime, pageable)
+      val referrals: MutableList<ReferralEntity> = referralPage.content
+
+      referrals.forEach { referral ->
+        val referralId = referral.id!!
+        val message = getDomainEventMessage(referralId, referral.crn)
+        log.info("Publishing ${ACP_M_AND_D_REFERRAL_DETAILS_UPDATED.value} event for referralId: $referralId")
+        publishEvent(message, referralId)
```
raby: The original publishEvent call on this line was not removed when the try-catch was added below, so every referral now publishes the domain event twice, and a failure here still re-throws and aborts the loop (the original issue is still present).

remove line 57?

pfcairns (dismissed):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJob.kt
```diff
@@ -0,0 +1,124 @@
...
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @Scheduled(cron = "\${app.scheduling.referral-details-updated.cron}")
+  fun process() {
+    val now = clock.instant()
+    log.info("Referral details updated Job started: at $now")
+    var totalUpdated = 0
+    val cutoffTime = LocalDateTime.now(clock).minusSeconds(cutoff)
+    val startingPage = 1
```
raby: `startingPage` is assigned once to initialise `currentPage` and never referenced again I think? Simplify to ` var currentPage = 1` or does have another use?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJob.kt
miroslav25a: good spot. comment addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/job/ReferralDetailsUpdatedJob.kt
miroslav25a: comment addressed

raby (approved):
Looks great

PR #787, opened 2026-07-09
pass first facilitator name as officer and add treatment manager to notes

rgforsyth:
pass first facilitator name as officer and add treatment manager to notes

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/client/nDeliusIntegrationApi/model/CreateAppointmentRequestTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/client/nDeliusIntegrationApi/model/CreateAppointmentRequestTest.kt
@@ -2,8 +2,11 @@ package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.cli
 
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/client/nDeliusIntegrationApi/model/CreateAppointmentRequestTest.kt
```diff
@@ -37,30 +40,113 @@ class CreateAppointmentRequestTest {
...
+  fun `toAppointment should produce null staff and team when session has no facilitators`() {
+    val session = buildSession(facilitators = emptyList())
+    val attendee = attendeeFor(session)
+
+    val appointment = attendee.toAppointment(UUID.randomUUID())
+
+    assertThat(appointment.staff).isNull()
+    assertThat(appointment.team).isNull()
+  }
+
```
raby, line 68: A minor comment: We could add a small test for a missing gap I thought of...


LLM help:
```
`   @Test
   fun `toAppointment should produce notes with only treatment manager when single facilitator`() {
       val treatmentManager = FacilitatorEntityFactory().withPersonName("TM Name").produce()
       val session = buildSession(facilitators = listOf(FacilitatorEntityFactory().produce()), treatmentManager = treatmentManager)
       val appointment = attendeeFor(session).toAppointment(UUID.randomUUID())
       assertThat(appointment.notes).isEqualTo("Treatment Manager: TM Name")
   }`
```

raby (commented):
Great work!

raby (approved):

PR #794, opened 2026-07-13
[APG-2487] add a lock mechanism for scheduled jobs

miroslav25a:


src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/sar/SarContractIntegrationTest.kt
```diff
@@ -120,7 +120,7 @@ class SarContractIntegrationTest :
       expectedApiResponsePath = "/sar/sar-api-response.json",
       expectedRenderResultPath = "/sar/sar-expected-render-result.html",
       attachmentsExpected = false,
-      expectedFlywaySchemaVersion = "111",
+      expectedFlywaySchemaVersion = "112",
```
raby, line 123: `SarContractIntegrationTest` hard-codes `expectedFlywaySchemaVersion = "112"` as a constructor arg, but the `hmpps-subject-access-request-test-support` library may also read `hmpps.sar.tests.expected-flyway-schema-version` from YAML independently. If both code paths are active, the mismatched value of `108` here could cause a test failure? Should` src/test/resources/application-test.yml `line 67 be updated to 112 to be safe?

raby (commented):
Overall the implementation is solid , great work. ShedLock with usingDbTime(), @Profile("!test") guard, and configurable lock durations are all correct. nice. Just one comment on sar tests. Thanks

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/sar/SarContractIntegrationTest.kt
miroslav25a: Thank you, comment addressed.

pfcairns (approved):
LGTM.

PR #796, opened 2026-07-14
APG-2401 Enable `dataRefreshEnabled` flag in production and explicitly set it to `false` in preprod and dev configurations.

pfcairns:


helm_deploy/values-preprod.yaml
```diff
@@ -23,3 +23,5 @@ generic-service:
 # See https://user-guide.cloud-platform.service.justice.gov.uk/documentation/monitoring-an-app/how-to-create-alarms.html#creating-your-own-custom-alerts
 generic-prometheus-alerts:
   alertSeverity: hmpps-accredited-programmes-manage-and-deliver-alerts-nonprod
+
+dataRefreshEnabled: false
```
Jack-Dempster, line 27: I know this will be right but just for my knowledge what is this enabling specifically? Is it like which env to pull from as I can see the prod one is true but the refresh will happen on pre prod right?

Jack-Dempster (approved):

helm_deploy/values-preprod.yaml
pfcairns: It's the k8s cron job that copies the prod database to the pre-prod DB, and so has to run on a prod k8s pod. There is some CP config behind the scenes that allows the prod job pod to see the pre-prod DB.

PR #801, opened 2026-07-16
APG-2401 Update cron job to use `postgres:17` for production-to-preprod data refresh.

pfcairns:
Upgrading docker image version due to job failure with the following errror message in the logs: 

```
pg_dump: error: aborting because of server version mismatch
pg_dump: detail: server version: 17.9; pg_dump version: 16.14 (Debian 16.14-1.pgdg13+1)
```


Alison-Weatherston (approved):
LGTM

PR #802, opened 2026-07-16
Tidy up sentry logging and filter out un actionable errors

Jack-Dempster:


pfcairns (approved):
Looks good. I'm questioning why we have two exception handlers though? Appreciate it's not part of your code review, so not for now,  but I reckon they should probably be consolidated.

PR #812, opened 2026-07-22
no-ticket/allow-ignore-of-new-pni-assessment-fields

pfcairns:
Ensure that any additional new fields added to the get pNI score API endpoint does not break existing PNI score retrieval functionality.


Alison-Weatherston (approved):
LGTM

PR #817, opened 2026-07-23
APG-2453 : Add local script for generating groups

Jack-Dempster:


scripts/seed-data.sh
```diff
@@ -37,19 +41,34 @@ wait_for_api() {
...
+    local group_count="${2:-$DEFAULT_GROUP_COUNT}"
     wait_for_api
 
     echo "Seeding $count referrals..."
     response=$(curl -s -X POST "${API_BASE_URL}/dev/seed/referrals?count=${count}")
     echo "$response" | jq . 2>/dev/null || echo "$response"
+
+    echo ""
+    echo "Seeding $group_count groups..."
+    response=$(curl -s -X POST "${API_BASE_URL}/dev/seed/groups?count=${group_count}")
```
raby: With `set -e (`L6) and no `--fail`, if the API returns a 4xx/5xx, curl exits 0 and the script happily continues. Teardown would look "complete" after a failed groups delete, then proceed to delete referrals with orphaned memberships. Add `--fail `to all four curl calls in` seed()` and `teardown()`?


```
# Current — a 500 from the API prints the error body but script continues
response=$(curl -s -X POST "${API_BASE_URL}/dev/seed/referrals?count=${count}")

# Suggested
response=$(curl -s --fail -X POST "${API_BASE_URL}/dev/seed/referrals?count=${count}")
```

```
# Current — a 500 from the API prints the error body but script continues
response=$(curl -s -X POST "${API_BASE_URL}/dev/seed/groups?count=${group_count}")

# Suggested
response=$(curl -s --fail -X POST "${API_BASE_URL}/dev/seed/groups?count=${group_count}")
```

Jack-Dempster: have added this now with the fail

niaevans9 (approved):
lgtm

PR #819, opened 2026-07-27
APG-2458 Add models and API client methods for licence conditions and requirements

pfcairns:
* Added get requirements and get licence conditions client code and model objects
* Added basic tests to verify functionality


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/client/nDeliusIntegrationApi/model/CodedValue.kt
```diff
@@ -0,0 +1,3 @@
+package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.nDeliusIntegrationApi.model
+
+data class CodedValue(val code: String, val description: String)
```
Jack-Dempster: I think we already have a class called `CodeDescription` that we've been using for this stuff. If we want this separate that's fine

pfcairns: No, that's fair. No point re-inventing the wheel.

Jack-Dempster (approved):
👍

PR #823, opened 2026-07-30
Add admin API to resolve invalid event numbers

pfcairns:
This pull request introduces a new admin API endpoint that enables resolving invalid event numbers. 

#### Changes Included:
- Added the admin API endpoint for resolving invalid event numbers.
- Updated service implementation to handle new nDelius client calls
- Associated backend logic to handle resolutions efficiently.
- Added JUnit and integration tests
- Added testing factories for new ndelius api responses model objects
- Added test wiremock stubs for new ndelius api calls


docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
     image: postgres:18
     networks:
       - hmpps
-    restart: always
```
pfcairns, line 26: Every morning when my laptop starts up and Docker desktop starts up, I have to stop these containers. This will prevent that from occurring.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AdminController.kt
```diff
@@ -214,4 +217,50 @@ class AdminController(
...
+        description = "Unauthorized",
+        content = [Content(schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+    security = [SecurityRequirement(name = "bearerAuth")],
+  )
+  @PutMapping("/admin/resolve-referral-event-numbers")
+  fun resolveReferralEventNumbers(): ResponseEntity<Void> {
+    log.info("Received request to resolve referral event numbers")
+    val referralsWithInvalidEventNumbers = referralService.findReferralsWithInvalidEventNumbers()
```
miroslav25a: Maybe it would be a good idea to move the lines 250 to 263 into service layer (referralEventNumberResolverService) as this looks more like a business logic? And as the status ResponseEntity.ok().build() is returned on both occasions we can just have one at the end.

pfcairns: Yes, that's a fair comment. Will address.

miroslav25a (dismissed):

miroslav25a (approved):

PR #827, opened 2026-07-31
System text updated to Accredited Programmes service automated update

Alison-Weatherston:
Update 'SYSTEM' to 'Accredited Programmes service automated update' in Status history. e.g. https://accredited-programmes-manage-and-deliver-dev.hmpps.service.justice.gov.uk/referral/f1fd3bc3-bdfb-4807-b6ad-eb66fd28eb57/status-history

Screenshot 2026-07-31 at 10 36 15

The only change to the 'src/test/resources/sar/sar-api-response.json' file:

Before: "createdBy":"SYSTEM"
After: "createdBy":"Accredited Programmes service automated update"


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/CohortService.kt
```diff
@@ -70,7 +70,7 @@ class CohortService(
 
   fun hasOverriddenCohort(referralId: UUID): Boolean {
     referralCohortHistoryRepository.findTopByReferralIdOrderByCreatedAtDesc(referralId)?.let {
-      return it.createdBy != "SYSTEM"
+      return it.createdBy != "SYSTEM" && it.createdBy != "Accredited Programmes service automated update"
```
Jack-Dempster: We have a constant in the code that is `ACCREDITED_PROGRAMMES_AUTOMATED_UPDATE` but it looks like its a different value than the one here but only slightly different. 

It would be good to have this as constant but can we check if these 2 contents should be the same?

Alison-Weatherston: @Jack-Dempster Flora confirmed that 'ACCREDITED_PROGRAMMES_AUTOMATED_UPDATE' is fine rather than introducing the slightly different name. So good spot - thank you

Jack-Dempster (approved):
👍

PR #831, opened 2026-08-03
APG-2458 Prevent duplicate referrals through validation.

pfcairns:
### Description

- The pull request introduces validation logic in `resolveReferralEventNumbers` to ensure duplicate referrals are prevented, which have occurred in testing due to the unique constraint : `uq_referral_crn_event_id_sourced_from`.  

- The licence/requirement data coming back from nDelius may inadvertently attempt to update referral data to a combination that already exists. 

### Changes Made

- Added a validation mechanism to check and avoid duplicate referrals in `resolveReferralEventNumbers` by performing an upfront test to verify that an existing referral with the "new" details doesn't already exist.
- Added an integration test to verify this scenario

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
@@ -73,6 +73,10 @@ class ReferralEventNumberResolverService(
       it.subCategory?.code == LICENCE_CONDITION_BUILDING_CHOICES_SUBCATEGORY_CODE &&
         it.subCategory.description == BUILDING_CHOICES_SUBCATEGORY_DESCRIPTION
     }?.let {
+      if (duplicateReferralDetailsExists(referral, it.id.toString())) {
+        log.warn("Existing referral details found for crn: ${referral.crn}, event_id: $it.id.toString(), sourced_from: ${referral.sourcedFrom}")
+        return false
+      }
       referral.eventNumber = it.eventNumber.toInt()
       referral.eventId = it.id.toString()
       referralRepository.save(referral)
@@ -90,6 +94,10 @@ class ReferralEventNumberResolverService(
         it.subCategory.description == BUILDING_CHOICES_SUBCATEGORY_DESCRIPTION
     }
     buildingChoicesRequirement?.let {
+      if (duplicateReferralDetailsExists(referral, it.id.toString())) {
+        log.warn("Existing referral details found for crn: ${referral.crn}, event_id: $it.id.toString(), sourced_from: ${referral.sourcedFrom}")
+        return false
+      }
       referral.eventNumber = it.eventNumber.toInt()
       referral.eventId = it.id.toString()
       referralRepository.save(referral)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
```diff
@@ -73,6 +73,10 @@ class ReferralEventNumberResolverService(
       it.subCategory?.code == LICENCE_CONDITION_BUILDING_CHOICES_SUBCATEGORY_CODE &&
         it.subCategory.description == BUILDING_CHOICES_SUBCATEGORY_DESCRIPTION
     }?.let {
+      if (duplicateReferralDetailsExists(referral, it.id.toString())) {
+        log.warn("Existing referral details found for crn: ${referral.crn}, event_id: $it.id.toString(), sourced_from: ${referral.sourcedFrom}")
```
raby: In Kotlin, `$it `in a string template interpolates the whole it object (`calls it.toString()`), then appends the literal text .id.toString(). This means the log output will be wrong - you'll get the full object representation, not the id.

 example fix:
`log.warn("...event_id: ${it.id}, sourced_from: ${referral.sourcedFrom}")`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
```diff
@@ -99,6 +107,16 @@ class ReferralEventNumberResolverService(
     return false
   }
 
+  // A unique key on the referral table exists to prevent duplicate referrals being created for a given crn, event_id and sourced_from
+  fun duplicateReferralDetailsExists(
```
raby: This is an internal implementation detail - not part of the service's public contract. It could be private unless it's intentionally exposed for direct unit testing? (there's no such unit test here; the integration test goes through the endpoint).

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AdminControllerIntegrationTest.kt
```diff
@@ -461,4 +461,52 @@ class AdminControllerIntegrationTest : IntegrationTestBase() {
     assertThat(updatedReferral3?.eventNumber).isEqualTo(5)
     assertThat(updatedReferral3?.eventId).isEqualTo("2500910296")
   }
+
+  @Test
+  fun `should not update referral event number if a duplicate referral already exists`() {
```
raby, line 466: The new test should not update referral event number if a duplicate referral already exists only exercises the REQUIREMENT branch. The resolveLicenceCondition path has the same duplicate-check logic (and the same string template bug) but has no equivalent test. A second test for LICENCE_CONDITION would give full coverage.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
pfcairns: Good spot, that should've been bracketed!

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralEventNumberResolverService.kt
pfcairns: Yes, agreed. Should be private.

src/test/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/AdminControllerIntegrationTest.kt
pfcairns: I don't think this is necessary as the duplicate referral already exists functionality is the same across both the requirements and licence conditions branch, so am happy that it works as expected.

raby (approved):
Looks good

PR #832, opened 2026-08-04
APG-2100 Add Capitalised midday

Alison-Weatherston:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralService.kt
```diff
@@ -734,6 +734,7 @@ class ReferralService(
             date = session.startsAt.format(DateTimeFormatter.ofPattern("d MMMM yyyy")),
             unformattedDate = session.startsAt,
             time = "${formatTimeForUiDisplay(session.startsAt.toLocalTime())} to ${formatTimeForUiDisplay(session.endsAt.toLocalTime())}",
+            timeWithCapitalisedMidday = "${formatTimeForUiDisplay(session.startsAt.toLocalTime(), capitaliseSpecialTimes = true)} to ${formatTimeForUiDisplay(session.endsAt.toLocalTime(), capitaliseSpecialTimes = true)}",
```
raby: The time formatting is inlined here while `ProgrammeGroupService` has a private `formatTimeOfSession(start, end, capitaliseSpecialTimes)` helper doing exactly this. Since it's private it can't be reused, but now both fields are formatted twice in a row with duplicated logic. Could be worth extracting `formatTimeOfSession` to a top-level function in `FormatTimeForUiDisplay.kt` so both services can share it - otherwise this pattern will drift if the formatting logic ever changes.

 FormatTimeForUiDisplay.kt:
```

> fun formatTimeOfSession(startTime: LocalTime, endTime: LocalTime, capitaliseSpecialTimes: Boolean = false): String {
>   val formattedStartTime = formatTimeForUiDisplay(startTime, capitaliseSpecialTimes)
>   val formattedEndTime = formatTimeForUiDisplay(endTime, capitaliseSpecialTimes)
>   return "$formattedStartTime to $formattedEndTime"
> }
```

Then ReferralService lines 736–737 become:
```
> time = formatTimeOfSession(session.startsAt.toLocalTime(), session.endsAt.toLocalTime()),
> timeWithCapitalisedMidday = formatTimeOfSession(session.startsAt.toLocalTime(), session.endsAt.toLocalTime(), capitaliseSpecialTimes = true),
``` And the private `formatTimeOfSession` in `ProgrammeGroupService` (line 616) could be deleted.

Alison-Weatherston: @raby 
Comments addressed - Thank you

raby: Thank you

raby (approved):
Looks good!

PR #837, opened 2026-08-05
APG-2563 When editing a session set facilitator rather than treatment manager

Jack-Dempster:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/SessionEntity.kt
```diff
@@ -124,3 +125,12 @@ class SessionEntity(
...
 }
+
+fun SessionEntity.primaryFacilitator(): FacilitatorEntity {
+  val byName = compareBy<SessionFacilitatorEntity>({ it.facilitator.personName }, { it.facilitatorCode })
+  val sessionFacilitator =
+    sessionFacilitators.filter { it.facilitatorType == FacilitatorType.REGULAR_FACILITATOR }.minWithOrNull(byName)
+      ?: sessionFacilitators.minWithOrNull(byName)
+  return sessionFacilitator?.facilitator
+    ?: programmeGroup.treatmentManager!!
+}
```
raby, line 138: `treatmentManager!! `will throw an unguarded `NullPointerException` if `sessionFacilitators` is empty and `treatmentManager` is null. There are four call sites across `ScheduleService` and `SessionService` with no equivalent branching check before calling `primaryFacilitator()`, so fixing it here is the cleanest approach. We could split into a null-safe chain with an explicit throw:

example

```
 return sessionFacilitator?.facilitator
   ?: programmeGroup.treatmentManager
   ?: throw IllegalStateException("No facilitator or treatment manager found for session: $id")
```

raby (commented):
Good work just one minor comment

pfcairns (dismissed):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/entity/SessionEntity.kt
Jack-Dempster: I did have this originally but took it out as from the UI we require a treatment manager. I'm gonna put this back in though as a failsafe as it seems silly not to add. Thanks for pointing out

raby (approved):

PR #838, opened 2026-08-05
Make feature flag for LAO access check environment-specific and update related configurations and documentation.

pfcairns:
Activates LAO access check in API for `dev` and `preprod`.


miroslav25a (approved):

raby (commented):
Good docs!

PR #839, opened 2026-08-06
Organise session controller tests into nested classes

Jack-Dempster:
Follow the existing pattern of sorting large test classes into inner classes by test endpoint.

No test cases have been changed as part of this pr


raby (approved):
Great idea! Clean

PR #842, opened 2026-08-10
Apg 2539/display username on status update

miroslav25a:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
```diff
@@ -616,14 +620,26 @@ class SessionService(
     UAAB -> Option("No - did not attend", null, outcome.code.name)
   }
 
+  private fun getUserByUsername(username: String): User? {
```
raby: Nit/Suggestion no blocker good pr: Both `ReferralService` and `SessionService` have an identical private `getUserByUsername` wrapper (same try/catch + UNKNOWN_USER_USERNAME branch). We could extract to a `getUserByUsernameOrNull` method on `UserService` to keep the logic in one place - low effort and prevents the two copies from silently diverging if the guard conditions change later.

raby (dismissed):
Good work

pfcairns (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/SessionService.kt
miroslav25a: comment addressed

raby (approved):
Good fix

PR #845, opened 2026-08-12
APG-2574 : Add LAO check on session details page

Jack-Dempster:
This PR returns LAO status on the get session details endpoints

It also adds an LAO service class which can be used across the service and feature flagged in one place for ease of use

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/LimitedAccessResolverService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/LimitedAccessResolverService.kt
@@ -1,0 +1,45 @@
+package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.service
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.ClientResult
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.probationAccessControlApi.ProbationAccessControlApiClient
+
+@Component
+class LimitedAccessResolverService(
+  private val probationAccessControlApiClient: ProbationAccessControlApiClient,
+  private val userService: UserService,
+  @Value("\${app.features.lao-badge-enabled}")
+  private val badgeEnabled: Boolean,
+  @Value("\${app.features.lao-view-restriction-enabled}")
+  private val restrictionEnabled: Boolean,
+) {
+  data class Access(
+    val lao: Boolean,
+    val isExcluded: Boolean,
+  )
+
+  fun resolve(username: String, crns: List<String>): Map<String, Access> {
+    val distinctCrns = crns.distinct()
+    if (distinctCrns.isEmpty()) return emptyMap()
+
+    val laoByCrn = if (badgeEnabled) distinctCrns.associateWith(::isLao) else emptyMap()
+    val accessibleCrns = if (badgeEnabled || restrictionEnabled) {
+      userService.getAccessibleOffenders(username, distinctCrns)
+    } else {
+      distinctCrns.toSet()
+    }
+
+    return distinctCrns.associateWith { crn ->
+      Access(
+        lao = laoByCrn[crn] ?: false,
+        isExcluded = crn !in accessibleCrns,
+      )
+    }
+  }
+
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/LimitedAccessResolverService.kt
```diff
@@ -0,0 +1,45 @@
+package uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.service
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.ClientResult
+import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.probationAccessControlApi.ProbationAccessControlApiClient
+
+@Component
```
pfcairns: Could be `@Service`

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/LimitedAccessResolverService.kt
```diff
@@ -0,0 +1,45 @@
...
+class LimitedAccessResolverService(
+  private val probationAccessControlApiClient: ProbationAccessControlApiClient,
+  private val userService: UserService,
+  @Value("\${app.features.lao-badge-enabled}")
```
pfcairns: I think the `lao-badge-enabled` feature flag is a duplicate of the `lao-access-check-enabled` flag. The wording feels a bit UI rather than API centric also.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/LimitedAccessResolverService.kt
```diff
@@ -0,0 +1,45 @@
...
+  private val badgeEnabled: Boolean,
+  @Value("\${app.features.lao-view-restriction-enabled}")
```
pfcairns: I've just created something similar for my PR called `exclusion-access-check-enabled` which I think is similar in intention to `lao-view-restriction-enabled` and is a bit more API rather than UI centric.

src/main/resources/application.yml
```diff
@@ -19,6 +19,8 @@ sentry:
 app:
   features:
     lao-access-check-enabled: ${LAO_ACCESS_CHECK_ENABLED:false}
+    lao-badge-enabled: ${LAO_BADGE_ENABLED:false}
```
pfcairns: These would need equivalent per environment helm config file definitions, as the deployment will probably fail without them:

- values-dev.yaml
- values-preprod.yaml
- values-prod.yaml

pfcairns (commented):
Good work.  A few minor comments, and potential clashes.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/LimitedAccessResolverService.kt
Jack-Dempster: have updated this to match since yeah I think its duplicated

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/LimitedAccessResolverService.kt
```diff
@@ -0,0 +1,45 @@
...
+  data class Access(
+    val lao: Boolean,
+    val isExcluded: Boolean,
+  )
+
+  fun resolve(username: String, crns: List<String>): Map<String, Access> {
+    val distinctCrns = crns.distinct()
+    if (distinctCrns.isEmpty()) return emptyMap()
+
+    val laoByCrn = if (laoAccessCheckEnabled) distinctCrns.associateWith(::isLao) else emptyMap()
```
raby, line 26: This calls `getCaseAccessByCrn()` once per CRN in series. For a session with many attendees this means N separate HTTP calls to Probation Access Control before the response returns. Not a blocker for current session sizes. Just being aware.

raby (dismissed):

pfcairns (approved):
Thanks for making the updates

src/main/resources/application.yml
Jack-Dempster: have addd these values now and updated it to the same as your pr

PR #846, opened 2026-08-12
APG-2578 Enable exclusion access check feature flag

pfcairns:
### Summary
- Add feature flag for exclusion access checks, which is turned on in dev and pre-prod, off in prod
- Refactor `isLao` and `isExlcuded` methods to additionally validate whether a restriction/exclusion is "current"
- Refactor user access code to UserService
- Added helm and yaml config for new feature flag
- Refactored getPersonalDetails endpoint to use no auth version of method, as auth should be handled by the visible referrals on case list.
- Updated related tests and configurations to support the new changes.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReferralController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReferralController.kt
@@ -136,7 +136,7 @@ class ReferralController(
 
     referralEventNumberResolverService.resolveIfEventNumberIsZero(referral)
 
-    return userService.getPersonalDetailsByIdentifier(referral.crn).let {
+    return userService.getPersonalDetailsWithoutAuthentication(referral.crn).let {
       ResponseEntity.ok(it.toModel(referral.setting))
     }
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralCaseListItemService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralCaseListItemService.kt
@@ -14,8 +14,6 @@ import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.caseList.StatusFilterValues
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.caseList.toApi
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.api.model.programmeGroup.ProgrammeGroupCohort
-import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.ClientResult
-import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.probationAccessControlApi.ProbationAccessControlApiClient
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.client.probationAccessControlApi.model.AllCaseAccess
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.entity.ReferralCaseListItemViewEntity
 import uk.gov.justice.digital.hmpps.accreditedprogrammesmanageanddeliverapi.repository.ReferralCaseListItemRepository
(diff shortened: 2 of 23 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReferralController.kt
```diff
@@ -136,7 +136,7 @@ class ReferralController(
 
     referralEventNumberResolverService.resolveIfEventNumberIsZero(referral)
 
-    return userService.getPersonalDetailsByIdentifier(referral.crn).let {
+    return userService.getPersonalDetailsWithoutAuthentication(referral.crn).let {
```
Jack-Dempster, line 139: is this intentional? does this not mean we aren't checking the lao on this page?

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralCaseListItemService.kt
```diff
@@ -145,35 +154,36 @@ class ReferralCaseListItemService(
...
-    is ClientResult.Success -> response.body
-    is ClientResult.Failure -> {
-      val exception = response.toException()
-      log.error("Failed to retrieve LAO case access for CRN $crn: ${response.getErrorMessage()}", exception)
-      throw response.toException()
-    }
+  private fun isExcludedByUsername(crn: String, username: String, caseAccessByCrn: Map<String, AllCaseAccess>?): Boolean {
+    val now = OffsetDateTime.now()
+    val caseAccess = caseAccessByCrn?.get(crn) ?: return false
+    return caseAccess.excludedFrom.any { it.username == username && (it.until == null || it.until > now) }
```
Jack-Dempster, line 186: I think we probably need to extract this logic into into it's own service which i've had a first stab at in my pr. Not for this PR t change but I think we may need to revisit or we'll have multiple implementations across the service

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/UserService.kt
```diff
@@ -196,4 +199,13 @@ class UserService(
       emptyList()
     }
   }
+
+  fun getCaseAccessByCrn(crn: String): AllCaseAccess = when (val response = probationAccessControlApiClient.getCaseAccessByCrn(crn)) {
+    is ClientResult.Success -> response.body
+    is ClientResult.Failure -> {
+      val exception = response.toException()
+      log.error("Failed to retrieve LAO case access for CRN $crn: ${response.getErrorMessage()}", exception)
+      throw response.toException()
```
Jack-Dempster:
```suggestion
      throw exception
```

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/controller/ReferralController.kt
pfcairns: Yes, the logic being the case list should now only show referrals that the user is allowed to see. This validation is currently firing randomly and logging users out as a result.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/UserService.kt
pfcairns: Thanks, good spot.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralCaseListItemService.kt
```diff
@@ -145,35 +154,36 @@ class ReferralCaseListItemService(
...
     return PageImpl(caseListReferrals.content, pageable, totalAllowedCount)
   }
 
-  private fun isLao(crn: String, caseAccessByCrn: Map<String, AllCaseAccess>?): Boolean = (caseAccessByCrn?.get(crn)?.excludedFrom?.isNotEmpty() ?: false) || (caseAccessByCrn?.get(crn)?.restrictedTo?.isNotEmpty()) ?: false
+  private fun isLao(crn: String, caseAccessByCrn: Map<String, AllCaseAccess>?): Boolean {
+    val now = OffsetDateTime.now()
+    val caseAccess = caseAccessByCrn?.get(crn) ?: return false
 
-  private fun isExcludedByUsername(crn: String, username: String, caseAccessByCrn: Map<String, AllCaseAccess>?): Boolean = caseAccessByCrn?.get(crn)?.excludedFrom?.any { it.username == username } ?: false
+    return caseAccess.restrictedTo.any { it.until == null || it.until > now } ||
```
raby, line 179: I think both methods` isLao()` and `isExcludedByUsername() `check until but never check `since`. A restriction/exclusion where `since` is in the future will incorrectly be treated as already active. The test helpers set `since = OffsetDateTime.now().minusDays(1) `by default, implying since maybe semantically meaningful. 

```
// ❌ Current — only checks `until`
return caseAccess.restrictedTo.any { it.until == null || it.until > now }

// ✅ Should also check `since`
return caseAccess.restrictedTo.any { it.since <= now && (it.until == null || it.until > now) }
```

Jack-Dempster (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ReferralCaseListItemService.kt
pfcairns: Good spot, have updated.

raby (approved):
looks good

PR #847, opened 2026-08-12
[APG-2581] add session with restricted access participants

miroslav25a:


src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
```diff
@@ -247,6 +253,53 @@ class ProgrammeGroupService(
...
+      ?: throw NotFoundException("Module with id $moduleId not found")
+
+    val suggestedDate = scheduleService.getNextSlotDate(groupId, moduleId)
+
+    val username = authenticationUtils.getUsername()
+    val userRegion = userService.getUserRegions(username).firstOrNull()
+      ?: throw NotFoundException("Region for username $username not found")
+    val facilitators = regionService.getTeamMembersForPdu(userRegion.code)
+    val memberships = programmeGroupMembershipService.getActiveGroupMemberships(groupId)
+    val groupMembers = memberships.map { membership ->
```
raby, line 270: The loop makes one HTTP call to the probation access control API per group member. A group with 10 participants = 10 serial HTTP calls. The same service already has the correct pattern elsewhere (e.g. group allocations ~line 450)

We could pre-fetch all CRNs first, then derive flags from the map? Maybe I am missing something too with context understanding, please dismiss if not valid.

```
val memberships = programmeGroupMembershipService.getActiveGroupMemberships(groupId)
val crnAccessMap = if (laoAccessCheckEnabled) {
    memberships.map { it.referral.crn }.distinct()
        .associateWith { limitedAccessOffenderService.getCaseAccessByCrn(it) }
} else null

val groupMembers = memberships.map { membership ->
    val crn = membership.referral.crn
    GroupMember(
        name = membership.referral.personName,
        crn = crn,
        referralId = membership.referral.id!!,
        isLimitedAccessOffender = limitedAccessOffenderService.isLimitedAccessOffender(crn, crnAccessMap),
        isExcluded = limitedAccessOffenderService.isExcludedByUsername(crn, username, crnAccessMap),
    )
```
}

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/LimitedAccessOffenderService.kt
```diff
@@ -0,0 +1,41 @@
...
+  val log = LoggerFactory.getLogger(this::class.java)
+
+  fun getLimitedAccessOffenderByCrn(crn: String): Boolean = when (val response = probationAccessControlApiClient.getCaseAccessByCrn(crn)) {
+    is ClientResult.Success -> response.body.excludedFrom.isNotEmpty() || response.body.restrictedTo.isNotEmpty()
+    is ClientResult.Failure -> false
+  }
+
+  fun isLimitedAccessOffender(crn: String, caseAccessByCrn: Map<String, AllCaseAccess>?): Boolean = (
+    caseAccessByCrn?.get(crn)?.excludedFrom?.isNotEmpty()
+      ?: false
```
raby: The trailing `?: false` binds to the entire OR expression (not just `restrictedTo?.isNotEmpty()`), so if `restrictedTo` is non-null but `isNotEmpty()` returns a real value, this still works ...however if `restrictedTo` is non-null and the whole expression evaluates to a `Boolean?` from the left side, the precedence is wrong. More practically, if you ever get a null from the safe-call chain on the left side, the right-hand `?: false` catches the whole thing, masking potential real values. Could be:?

```
(caseAccessByCrn?.get(crn)?.excludedFrom?.isNotEmpty() ?: false) ||
(caseAccessByCrn?.get(crn)?.restrictedTo?.isNotEmpty() ?: false)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/model/programmeGroup/GroupMember.kt
```diff
@@ -6,14 +6,20 @@ import java.util.UUID
...
   @get:JsonProperty("referralId", required = true)
-  @Schema(description = "The UUID of the referral for this group member")
+  @get:Schema(description = "The UUID of the referral for this group member")
   val referralId: UUID,
+
+  @get:Schema(description = "The boolean value of whether the group member has Limited Access Offender (LAO) status")
+  val isLimitedAccessOffender: Boolean? = false,
+
+  @get:Schema(description = "The boolean value of whether the group member details are excluded from viewing by the logged-in username")
+  val isExcluded: Boolean? = false,
```
raby: These are never null in practice? Could be non nullable. Maybe a reason why chosen to be.

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/ProgrammeGroupService.kt
miroslav25a: good point. comment addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/service/LimitedAccessOffenderService.kt
miroslav25a: class removed

src/main/kotlin/uk/gov/justice/digital/hmpps/accreditedprogrammesmanageanddeliverapi/api/model/programmeGroup/GroupMember.kt
miroslav25a: there is now. please check ProgrammeGroupService lines 271 and 283

raby (approved):

